PR #15, opened 2015-01-11
documentation clarification for linechart data

jefffriesen:
It looks like from the code that the intended data structure for passing in a series of line charts is an object. I've updated the example and added a `propType`.


esbullington:
Thanks, Jeff.  I missed that when I merged that pull request, and we don't have tests for the example code.  We're currently not able to unit test the example code at this point, because of the particular (convoluted) format I chose for the examples.  But I'm hoping to soon move to more comprehensive documentation with examples using the interactive [React application](http://react-bootstrap.github.io/components.html) that React-Bootstrap has created for their project's documentation.

PR #17, opened 2015-01-15
Add treemap component

yang-wei:


esbullington:
Looks great!  Thanks, @yang-wei

One small update: if you could please just change the default palette from `d3.scale.category20b()` to `d3.scale.category20c()` (which is what all the other charts that use d3's palettes are using), I'll go ahead and merge your PR immediately after that you push that update.

One possible additional feature would be to pass down the `textColor` and `fontSize` from the top-level component, thus allowing users to easily customize these values without changing the library code, but that's a minor change that we can always add later.

Thanks again!

yang-wei:
@esbullington You're welcome ! I just changed the color and thinking to implement zoomable treemap next =)  
Also, would like to make more component for this repo because I really find this is a good way to learn d3.js fast.

esbullington:
Perfect, thanks!  I'm going ahead and merging it now.  I plan on publishing another release to NPM in the next day or two, so your treemap will soon be available via NPM as well.

Also, feel free to propose another chart types! It would be great if we could eventually reach feature parity with nvd3.js or even Highcharts, so any of the charts types from those libraries would be welcome.

Regarding animation, I'm still thinking about how to handle that with React.  I'm thinking about making an animation mixin to hold state for animated components, but haven't decided 100%.  Could I ask you to please hold off on any pull requests with animation for the moment?  I'll add info to the CONTRIBUTING.md file as soon as I settle on an approach.

PR #33, opened 2015-01-23
Allow axes to go negative

berrytj:
This change preserves the linechart's behavior for data consisting of only positive values – the axes will still start at 0.  But for data with negative values, the axis will begin at the lowest negative value (see x-axis of image below).  What do you think?

I'm happy to add a test, but since the axes are not rendered by React, the ticks aren't found by the TestUtils methods.

screen shot 2015-01-23 at 11 21 24 am

### DIFF
--- a/src/areachart.jsx
+++ b/src/areachart.jsx
@@ -6,6 +6,7 @@ var common = require('./common');
 var Chart = common.Chart;
 var XAxis = common.XAxis;
 var YAxis = common.YAxis;
+var _ = require('lodash');
 
 var Area = React.createClass({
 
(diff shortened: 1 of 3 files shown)


esbullington:
Hey, thanks for this.  Yes, we do need to accommodate negative values.  Ideally, I was planning on creating something like [Highchart's `threshold`](http://jsfiddle.net/esbullington/vuq4wvdv/), which allows you to either visualize negative values as dipping below the x axis, or to visualize them as you're doing here, by starting the origin of either axis with the least negative value instead of at 0.

But I'll go ahead and merge it and plan on adding a `threshold` type feature at a later time, since I wasn't even planning on addressing this until after 0.2.0.

Thanks again!

berrytj:
Makes sense.  Thank you!

PR #42, opened 2015-01-25
Add Legend component

esbullington:


esbullington:
Hey @yang-wei  After thinking about in long and hard, and testing out both an SVG and an HTML-based legend, I decided to go with the HTML component.  Ideally, I would have preferred to use SVG, but SVG text is notoriously hard to work with.  In particular, SVG cannot wrap text, which turns out to be pretty important for legends with long series names.

I decided that since a) we already had the title as an HTML element and b) since the legend is really an integral part of the chart itself, it should be ok.

The downside to using an HTML-based component is that we don't get free zooming/resizing since it's not part of the core chart SVG element. However, looking at various online charts, few of them resize the legend and title to changes in the chart itself.  Plus, since it's HTML, we can easily resize using media queries, when the time comes.

What do you think?

yang-wei:
yes I was having same svg issue when I was developing [gh-top](https://github.com/yang-wei/gh-top). So in this case I think HTML will be fine. Go ahead and merge it, Sir !

esbullington:
Thanks for taking a look, @yang-wei Appreciate it!

Issue #28, opened 2015-01-22
v0.2.0


esbullington:
To do for v0.2.0:

1) Add maybe two more chart types: a simple gauge chart (esbullington) and  scatter (yang-wei) 
2) Improve documentation (esbullington working on it)
3) Add legends (esbullington working on it)
4) Get the code base cleaned up, and start using jshint (esbullington working on it)
5) Factor out the axes (almost done by esbullington in #27 )
6) Add some simple mouseover animations for the line, bar, and scatter charts  (yang-wei)

berrytj:
Hi @esbullington, I'm thinking of adding labels to the linechart – is this something you've already worked on?

berrytj:
Not sure if that's what you mean by "legend", above.

esbullington:
Hi @berrytj thanks for your input and contribution (I'll check your PR right now).  Yes, I'm working on both a [legend](http://bl.ocks.org/ZJONSSON/3918369) and [label](http://www.d3noob.org/2012/12/adding-axis-labels-to-d3js-graph.html) components right now.  Should have it up today or maybe tomorrow morning.

yang-wei:
hey @esbullington regarding to your todo 6) do you have any idea on mind? Refer to [nvd3](http://nvd3.org/examples/scatter.html) examples, "mouse over" will
 * change the color of dots to darker color
 * increase the size of dots
 * show extra informations on tooltips
Since we're only on early stage so do you think that we can implement the first 2 animation(color and size)

esbullington:
hey @yang-wei that's exactly what I have in mind: the first two things mouse over animations would be great for this release.

Tooltips are more complicated so we can save that for a later release.

yang-wei:
cool ! I'm willing to take over that =) @esbullington

esbullington:
That'd be great, @yang-wei  I'm finishing up the legends and also working on automating the build, release and npm publish process a bit more, as well as documentation.

I've restarted the legend several times because I've not been happy with the appearance.  But I think I found a look I like.

I also plan on finishing up the gauge chart before 0.2.0.

yang-wei:
@esbullington Hey regarding to the animation, I have question to ask. Because most of our `fill`(color) and `r`(radius) are properties, so one way I can think of is to set up `onMouseOver` event on them, when the event is trigger we'll pass it to the top parent (which i think is `<Chart />`) and re-render them using `setProps()`. 

[...]

PR #49, opened 2015-01-28
Add hover animation

yang-wei:
Please note that instead of setting `color` as state, I add another state `strokeWidth` to help animating. Also I use `shouldComponentUpdate` to decide if they should do the animation. Feel free to correct me if you think there is another better way ( or you prefer your approach stated in #28 ). I din't any test for the animation since we don't have test spies. If you think adding test for animation is necessary, I will try to add them by adding dependency like `sinon.js`. 

If you're cool with this kind of approach, I will add animation for other charts on this branch too. :bar_chart: :chart_with_upwards_trend:

### DIFF
--- a/src/scatterchart.jsx
+++ b/src/scatterchart.jsx
@@ -14,13 +14,32 @@ var Circle = React.createClass({
     cx: React.PropTypes.number,
     cy: React.PropTypes.number,
     r: React.PropTypes.number,
-    fill: React.PropTypes.string
+    fill: React.PropTypes.string,
+    stroke: React.PropTypes.string,
+    strokeWidth: React.PropTypes.number,
+    strokeOpacity: React.PropTypes.number,
+    hoverAnimation: React.PropTypes.bool
   },
 
   getDefaultProps: function() {
     return {
-      fill: '#1f77b4'
+      fill: '#1f77b4',
+      stroke: '#1f77b4',
+      strokeWidth: 0,
+      strokeOpacity: 0.5
     };
+  },
+
+  getInitialState: function() {
+    // state for animation usage
+    return {
+      circleRadius: this.props.r,
+      strokeWidth: this.props.strokeWidth
+    } 
+  },
+
+  shouldComponentUpdate: function(nextProps, nextState) {
+    return this.props.hoverAnimation; 
   },
 
   render: function() {
@@ -28,10 +47,29 @@ var Circle = React.createClass({
       <circle
         cx={this.props.cx}
         cy={this.props.cy}
-        r={this.props.r}
+        r={this.state.circleRadius}
         fill={this.props.fill}
+        stroke={this.props.fill}
+        strokeOpacity={this.props.strokeOpacity}
+        strokeWidth={this.state.strokeWidth}
+        onMouseOver={this.animateCircle}
+        onMouseOut={this.restroreCircle}
       />
     );
+  },
+
+  animateCircle: function() {
+    this.setState({ 
+      circleRadius: this.state.circleRadius * ( 5 / 4 ),
+      strokeWidth: 3 / 2 
+    });
+  },
+
+  restroreCircle: function() {
+    this.setState({ 
+      circleRadius: this.state.circleRadius * ( 4 / 5 ),
+      strokeWidth: 0   
+    });
   }
 
 });
@@ -53,7 +91,7 @@ var DataSeries = React.createClass({
   render: function() {
 
     var circles = this.props.data.map(function(point, i) {
-      return (<Circle cx={this.props.xScale(point.x)} cy={this.props.yScale(point.y)} r={this.props.pointRadius} fill={this.props.color} key={this.props.seriesName + i} />);
+      return (<Circle cx={this.props.xScale(point.x)} cy={this.props.yScale(point.y)} r={this.props.pointRadius} fill={this.props.color} key={this.props.seriesName + i} hoverAnimation={this.props.hoverAnimation} />);
     }.bind(this));
 
     return (
@@ -79,7 +117,8 @@ var ScatterChart = React.createClass({
     axesColor: React.PropTypes.string,
     title: React.PropTypes.string,
     colors: React.PropTypes.func,
-    legend: React.PropTypes.bool
+    legend: React.PropTypes.bool,
+    hoverAnimation: React.PropTypes.bool,
   },
 
   getDefaultProps: function() {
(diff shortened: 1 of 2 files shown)


esbullington:
Hey @yang-wei thanks for the PR!  Sorry for not getting back earlier -- it's been a busy day.

So, I looked the code, and I describe a few issues below:

1) You wrote "...I use shouldComponentUpdate to decide if they should do the animation."  The problem with this approach will come up when we need to update properties in the `Circle` component, like when we do data frame animations.  If `hoverAnimation` is set to `false`, then no updates will work in the Circle component, because `shouldComponentUpdate` will always return false.  I'd recommend doing something like the approach in #28, which is to use a ternary expression when setting the event handler:

`onMouseOver={this.hoverAnimation ? this.animateCircle : null}`

Alternately, you could render different versions of the circle element depending on whether or not hoverAnimation is true or false, with one version setting the mouse event handlers, and the other without the handlers.

2) Little spelling error: `restroreCircle` -> `restoreCircle`

3) So regarding tests, it's actually pretty cool because React TestUtils lets us do things like simulate events on the virtual DOM.  So I can test your mouseover animation by doing this:

```
    var scatterchart = TestUtils.renderIntoDocument(
      <ScatterChart pointRadius={pointRadius} data={data} width={400} height={200} />
    );
    var circles = TestUtils.scryRenderedDOMComponentsWithTag(
      scatterchart, 'circle');
    var circleOne = circles[0];
    expect(circleOne.getDOMNode().r.baseVal.value).to.equal(pointRadius);
    TestUtils.Simulate.mouseOver(circleOne);
    var newPointRadius = pointRadius * (5 / 4);
    expect(circleOne.getDOMNode().r.baseVal.value).to.equal(newPointRadius);
```

So it would be great if you could add that to the scatterchart-tests.js file to test this addition.

4) Regarding the actual mouseover animation, I like the effect.  It's subtle, which is a good thing here.  I think you could also do something like darken/lighten the color by a certain percentage.  This would be easy in Less or SASS, but here we must [do it in Javascript](http://stackoverflow.com/a/13542669/1002047).

But this is not critical for now -- we can make this change later.

5) There's one more issue to address, which is how to detect nearby mouseevents.  RIght now, it only fires when the mouse is *right* over circle.  Ideally, we'd do something like MetricGraphics does [here](http://metricsgraphicsjs.org/examples.htm) (see "scatterplots" at bottom of page).  A common approach to detect nearby mouseover events is to use a [Voronoi-tessellated](http://bl.ocks.org/mbostock/8027835) overlay, in this case an svg `rect` overlay.  Using Voronoi tessellation is how most scatter charts I've seen pick up mouseover events that are not directly on top of the circles, but nearby.

However, this is not a huge issue, and I don't want to overwhelm you with feedback.  I'm perfectly happy to merge the PR once tests are added and the `shouldComponentUpdate` issue is changed.

Do you want to discuss this approach above and implement it yourself, or do you want to just add the tests and address the `shouldComponentUpdate` and then let me do it later?

For the other animations for the other charts, it'd be great if you could open an issue before each PR, so we can discuss what the animation approach (I'm open to new ideas, just would be good to discuss them first).

Thanks again for the PR -- I look forward to merging it.

yang-wei:
1), 2), 3) should be done =) Really thanks for the feedback. I never know that we can do something like this.
```
component.getDOMNode().r.baseVal.value
```

Regarding to issue 4), 5), I am willing to give them a try and send you a pr if everything go smoothly. I must say these 2 features are really a cool addition for animated graph. First of all, I will add them as a method in scatter chart component, if they are reusable(I think they should) then we can used them as mixins later.

Should I open another pull request for these 2 features ?

yang-wei:
Hey @esbullington I added the `shade` method and was trying to add test for the animation (color) but I can't access its `fill` property.
```
var circleOne = circles[0];
    expect(circleOne.getDOMNode().r.baseVal.value);  // get props r value
    expect(circleOne.getDOMNode().fill.baseVal.value); // undefined
```
I also try log the loop of `getDOMNode()` and dint see any value related to fill. 
```
var circleOne = circles[0];
var circleOneColor = circleOne.getDOMNode();
 for(var obj in circleOneColor) {
        console.log(obj);
 }
// Results
LOG: 'systemLanguage'
LOG: 'cx'
LOG: 'className'
LOG: 'requiredFeatures'
LOG: 'farthestViewportElement'
LOG: 'nearestViewportElement'
LOG: 'cy'
LOG: 'style'
LOG: 'transform'
LOG: 'r'
LOG: 'xmllang'
LOG: 'requiredExtensions'
LOG: 'xmlspace'
LOG: 'externalResourcesRequired'
LOG: 'viewportElement'
LOG: 'ownerSVGElement'
LOG: 'id'
...
```
This give me a lot of DOM related properties but not React component properties.
Do you have any idea?

esbullington:
Hey, you can access the fill value using: `circleOne.getDOMNode().getAttribute('fill')`.  Here is a [site](http://www.w3.org/TR/SVGTiny12/attributeTable.html) that gives a brief overview of the SVG Javascript API, and how to access normal attributes using the normal XML interface (as above) vs the special SVG DOM (like `baseVal.value`).

yang-wei:
I found something on the [spec](http://www.w3.org/TR/SVG/shapes.html#DOMInterfaces). SVG  `<circle>` element  only has 3 *readonly* attributes -  `cx`, `cy`, `r`. So I cannot do `getDOMNode().fill.baseVal.value`.

Instead I will just do something like this:
```
var circleOne = circles[0];
    expect(circleOne.props.r);  // get props r value
    expect(circleOne.props.fill); // get props fill value
```
What do you think? Sorry for spamming your feed hahaha

yang-wei:
Thank you. Using `getDOMNode()` is just what I wanted, thanks ! Will change it =)

esbullington:
You can access the fill value using `getDOMNode.getAttribute('fill')` but the method you are using is better.

yang-wei:
:+1:

esbullington:
Let me look over the whole thing again quick but I'm definitely ready to merge.  Looks good!

esbullington:
Thanks, this is great.  Do you want to have a shot at the Voronoi overlay to detect nearby mouse events for the scatter chart?

Also, the same approach (Voronoi) can be used to give the kind of animation [seen in line charts here](http://www.w3.org/TR/SVGTiny12/attributeTable.html), which I think is a nice kind of animation for line charts.

But please let me know if this is a busy time for you.  It can wait, or else I can do it.  You've already done a lot for this upcoming 0.2.0 release.

Thanks again!

yang-wei:
@esbullington I'm willing to look at that on this weekend. Thanks for the example !

PR #56, opened 2015-02-01
Initial implementation of veronoi to scatter chart

yang-wei:
Hey @esbullington I implemented voronoi in scatter chart. It is working but I need you to review this branch. This was a quick code so please ignore bad naming, no Proptype etc. 

There are something I am really not happy ( search for ugly hack in my code ). To implement voronoi, I have to concat all the data and add a new `color` property for each data(point). 
Have a look when you're free and let me know what do you think.

### DIFF
--- a/src/scatterchart.jsx
+++ b/src/scatterchart.jsx
@@ -8,7 +8,7 @@ var XAxis = common.XAxis;
 var YAxis = common.YAxis;
 var _ = require('lodash');
 
-var Circle = React.createClass({
+var Region = React.createClass({
 
   propTypes: {
     cx: React.PropTypes.number,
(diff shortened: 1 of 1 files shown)


esbullington:
Thanks, yang-wei, it works great despite the clever hack you had to do!  But you're right that it's a bit of an awkward workaround, not through any fault of your own.   We're definitely at the point where we need to figure out application-wide state, or our workarounds will become increasingly hacky.  You had to create that hack to overcome our lack of a way to pass around application state (the other option would have been to pass a callback around, but that gets old fast).

With this in mind, I went ahead and [pushed a branch](https://github.com/esbullington/react-d3/tree/voronoi) that uses EventEmitter to address this particular issue of passing around a bit of state, in this case which polygon path element has a mouseover event.  It allows us to keep the `DataSeries` as they are now (which I'd prefer since I want to continue to be able to pass in custom properties for each data series to a React component).

I used your Voronoi code to make a separate Voronoi component (in `common/voronoi`), to which I pass EventEmitter pubsub system, which notifies the `Circle` component when its corresponding Voronoi polygon has a mouseover event (I used the dataseries name + array index for a unique composite key to link together the circle and polygon).

I plan on switching this simple pubsub system to the Flux pattern soon -- I just need a little more time.

I may will still use a cursor to model the state of chart frames, but the Flux pattern will help us deal with these little bits of cross-cutting ephemeral state, and for various notifications.

Anyway, if you'd like to pull down this `voronoi` branch and adapt/polish it, and then use it to modify this PR (or make a new one), that'd be fine since a lot of that code was written by you.  I'm afraid I can't merge this PR as it is now since we need the `DataSeries` to remain modular and to correspond 1-to-1 with the data series that are passed in.  But it was a clever solution!

yang-wei:
Ok so we're using event here. I like that ! Just realised the test wasn't passed haha. Will open another pr based on your branch. Thanks for everything @esbullington . Will close this =)

PR #59, opened 2015-02-03
Clean up voronoi.jsx and make scatter chart's test pass

yang-wei:
Thanks @esbullington for factor out `voronoi.jsx`. This cleans up a lot.


esbullington:
Thanks, @yang-wei Much appreciated!

PR #76, opened 2015-02-07
Line chart voronoi

yang-wei:


yang-wei:
@esbullington Do you think that we need to make hover animation optional ? I just realised that it's not optional in `scatter chart` too.

esbullington:
Yeah, that would be great: `animation={true}`

esbullington:
@yang-wei I added some line animation to this, check it out in the `yang-wei-linechart-variation` branch.  Basically, I animate the lines now in addition to the circles.  I did need to make a small switch to the Voronoi component, which was to make `opacity` = 0, since I needed to put it in the foreground to prevent triggering `mouseout` and `mouseover` events every time I cross a line.

If you like, I'll push it to this branch, add the optional animation props, and we'll merge.

yang-wei:
Oh yeah I din't think of that. Great feature, just push it ! 

One small thing, we might have to change the event name (`animateCircle`) to something more general because we are animating other thing.

esbullington:
Thanks, @yang-wei I'll make these and other changes sometime tomorrow (had a busy weekend).  Thanks for your input on the data format question.

esbullington:
Hey, @yang-wei I merged your branch to one of my feature branches, which I am about to merge.  Nice work!

yang-wei:
:+1:

PR #78, opened 2015-02-07
Utiljs polish up and test

yang-wei:


esbullington:
Nice, very smart!  I keep forgetting about es6 string templates.  Merged.

PR #84, opened 2015-02-09
Fixes bug with xAsisClassSelect

martinRoss:
Need to selectAll on xAxisClassSelect (as you are doing for the YAxis). This selection only updates the first instance of xAxis but fails if you have multiples of the same chart type on a page.


frederickfogerty:
+1 (y)

esbullington:
Thanks, @martinRoss Merged.

PR #88, opened 2015-02-12
v0.2.0

esbullington:


esbullington:
Hey @yang-wei would you take a look at this v0.2.0 branch if you have time?  It's got some breaking changes, like the new data format, which is now standard across all the charts capable of handling multiple series.  Thanks!

frederickfogerty:
What happened to this? Seems like you remerged everything, somehow.

I'm not even mad

esbullington:
Yeah, I rebased it, since I merged a few PRs with master.  I'm squashing these down now.  Are you working from v0.2.0?

frederickfogerty:
I've got a local npm build running off v0.2.0, waiting for it to be released to the public npm registry.

esbullington:
@frederickfogerty Since you're working off of your own fork, you might want to check for a somewhat subtle bug in the `flattenData` function you contributed to, in this ternary expression:

    `accessors[dimension] ? accessors[dimension](item[dimension]) : item[dimension]`

The `item[dimension]` already accesses the object property before the accessor can even try, so I just changed the argument from `item[dimension`  to just `item` (I also changed spelling of `accesors` to `accessors`, so don't let that throw you off).  I missed it when I look at the PR, and only caught it when some of the charts weren't rendering properly.

I'm not going to try to squash commits because I don't want to squash other peoples' commits and trying to pick some from them all is a pain.  So viva repetitive commit lists!  I need to find out if there is a better way to rebase when a PR gets stale.

Thanks again for your help!

frederickfogerty:
Ah yes, I didn't pick that up, good spotting. 

Hopefully I can contribute more of my work in the future!

PR #90, opened 2015-02-12
Pass fontSize and textColor props down to Cell.

martinRoss:
These two props were not getting passed down to the Cell component.


esbullington:
thanks for that, @martinRoss

PR #91, opened 2015-02-12
Utils upgrade to work with any number of dimensions

frederickfogerty:
What do you think of this? This is backwards-compatible with current code and handles multiple dimensions.

It's also more resilient if either `xAccessor` or `yAccessor` isn't defined.


esbullington:
This looks good, thanks!  And needed, in view of our candlestick conversation.

Glad you saw the v0.2.0 branch I just pushed up earlier today.

PR #94, opened 2015-02-13
Could be useful to use this. First seen used by atom, and in many other packages since then

frederickfogerty:
https://github.com/atom/atom/blob/master/CONTRIBUTING.md

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -18,6 +18,29 @@ _By contributing code to React-d3, you are agreeing to release it under the MIT 
 
 ## SVG Performance Notes
 * Whenever possible, do transforms on groups of items (using `<g>`) as opposed to individual items.
- 
+
+## Git Commit Messages
+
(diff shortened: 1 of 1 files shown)


esbullington:
I appreciate your PR, but I'm just not a big fan of these emoji in commit messages.  That said, the Atom CONTRIBUTING.md has some excellent points about how to write good commit messages (ones that I don't always live up to, but aspire to). So if you remove the emoji section I'm all for merging this.

By the way, I like how you namespace your branches.  As we start to get more contributors and PRs, it's a good idea.  I'll try to start doing the same.

esbullington:
@frederickfogerty did you want to change this?  Please let me know, otherwise I'll go ahead and close it.

frederickfogerty:
I removed the emojis, so it's up to you whether you merge it or not

esbullington:
@frederickfogerty wow, you removed the emojis but then put one directly in the CONTRIBUTING?

PR #104, opened 2015-02-14
adding react-tools to make it easier to build

jmuel:
This seems to fix the dependencies so I can just do "npm install && npm run build" and it all works.

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,6 +30,7 @@
     "karma-phantomjs-launcher": "^0.1.4",
     "mocha": "^2.1.0",
     "nodemon": "^1.3.7",
+    "react-tools": "^0.12.2",
     "reactify": "^1.0.0",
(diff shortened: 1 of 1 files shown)


frederickfogerty:
+1

esbullington:
Thanks, I'll go ahead and merge this but I don't think it will address the issue you ran into until we change the two `jsx` references in the npm scripts to `./node_modules/react-tools/bin/jsx`.  If you'd like to make that change to this PR, that'd be great.  Otherwise, I'll be sure to add it next time I work with `package.json`.

Also, we have the `jsxhint` reference you might as well add and change to its local equivalent while you're at it (`./node_modules/jsxhint/bin/jsxhint`).

Thanks again.

esbullington:
Thanks, I'll merge it and make those changes myself.

PR #110, opened 2015-02-15
I'm doing a little cleanup on the build scripts. The most notable issue ...

jmuel:
...is that 'build' was changed to 'release' and the minified script wasn't updated to accomodate that. It was still running "npm run build" which doesn't do anything. I also updated clean so it removes the build/cjs directory since release will make the directory and populate it. Previously it would clean up only javascript files directly inside of dist/cjs/ and not, for example, dist/cjs/util/index.js or any of the .md that release copies in.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,6 @@ npm-debug.log
 
 # Build directory
 build/
+
+#intellij
+.idea/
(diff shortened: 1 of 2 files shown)


esbullington:
Thanks for catching that, much appreciated!

PR #113, opened 2015-02-16
Candlestick charts

rrag:
* Candlestick chart
* on mouse hover animation with Vornoli
* updates to utils.flatternData for conditions where yAccesor returns an object like `{open: ..., high: ..., low:..., close:... }`
* candle width is too small if larger time frame is used, so the examples use `dist/public/data/AAPL_ohlc.tsv` instead of `dist/public/data/AAPL_ohlc-full.tsv`


esbullington:
Hey @rrag this looks great.  I'll go over the code later today when I get a free block of time.  Looking forward to merging it!

esbullington:
Thanks @rrag I'm going ahead and merging it.  Nice job. 

One question: would you be OK if I cut out about 20% more of the data from the example data `AAPL_ohlc.tsv`?  On the tiny example charts, I think about 20% less example data makes your chart look better in that small space.  For the future, I'm trying to think of a way to nicely show example code for a full-width chart, which I think the candlestick and stacked area charts need to show up nicely.

Also, do we need the other ohlc file, the `AAPL_ohlc-full.tsv` file for anything?  We can always put it back in if we need it, right?

rrag:
Sure, go ahead and remove another 20% of the AAPL_ohlc.tsv. Please go ahead
and delete the -full.tsv, it is not used now. I just added it first and
then had to remove some data to get the chart to show up good. I should
have done it myself

PR #114, opened 2015-02-16
updating the demo link

jmuel:
The old demo link in the README is pointing to a page with out of date instructions. Changing the link to point to a more current demo page.


yang-wei:
Good one ! @jefffriesen

Issue #109, opened 2015-02-15
Alternate build script


rrag:
I had to use windows for a few days and the `npm run build` and `watch` do not work on the windows version since `mkdir -p` does not exist on windows cmd and the `mkdir` command returns an error when the directory already exists aborting the `npm run ...`. It spent too much time than I cared to on that problem and had to give up after some time.

I can create a gulpfile.js for the current build and send it over are you cool with that? or do you have other plans?

esbullington:
Hey @rrag I'm sorry you had to spend so much time on it.  jmuel and I made some improvements to the build script today but I didn't realize the `-p` flag on `mkdir` didn't work on Windows.  I was definitely aiming for cross-platform commands.

Ideally, I'd like to just use a `Makefile` but I know that there are cross-platform issues there, too. And if I'm accepting contributions from devs working on Windows machines, then they deserve to have a build tool that is easy for them to work with.

So yes, if you think Gulp is the best cross-platform answer to our problem, then I'll gratefully accept a PR for a gulpfile.js.  But let's please leave the current npm scripts in place until we're confident that the gulpfile has matched all of those scripts' current capabilities.

jmuel:
I actually prefer Gulp as well. Just using NPM scripts is nice when things are simple but even in this project so far we've had issues with globally installed react-tools and not maintaining aliases, both of which are hard problems to have with Gulp. I'd really appreciate this.

rrag:
why do we need this

`jsx --harmony -x jsx src build/cjs && jsx --harmony src build/cjs`

portion of the `release`?

Can I skip that step and just generate `react-d3.js` and `react-d3.min.js`

esbullington:
@rrag We need that portion of release because I need to compile the JSX files into pure JS files in their own package directory, which I then use to publish.  The directory is `build/cjs`.  The resulting build in that `build/cjs` directory is the definitive build that I publish to NPM. The react-d3.min.js is just for the JSFiddle, it's got React and d3 both packaged in (haven't been able to get a browserify standalone build to accept a global reference to dependencies, even though there should be a way).  Both the documentation charts that we publish to the website and the minified build for JSFiddle are just there for documentation and examples.

esbullington:
We don't want to publish the NPM package from anything like an aggregated react-d3.min.js that's not since we want people to be able to `require` the charts separately.  Plus that react-d3.min.js build is a standalone build not meant to be published to npm.

[...]

PR #118, opened 2015-02-17
add gulp build for the scripts in package.json

rrag:
gulp build script
There are some optimizations that can be done 

* split it into multiple files
* not load the plugins used by the current task

but this is a start so I can use my windows pc when I want to :)

gulp release does not do everything the npm run release does, needs review
gulp lint does not work as the gulp-jsxhint plugin is not complete, need an alternative or stick with npm run lint

execute gulp and see the list of tasks and their functions

Other notes:
Once the package.json scripts are retired some of the devDependencies can be eliminated.


rrag:
this is for #109

esbullington:
Took a closer look, and here are a couple of remarks that may clarify things:

1) I wasn't using es6ify, which depends on `traceur` es6 transpiler, which requires a number of shims for the transpiled code.  I was using react-tools itself to do the ES6 transpilation.  I've not used gulp, but looking at `gulp-react`, it appears you do this by passing the `harmony` flag to the `gulp-react` `react function, [as shown here](https://github.com/sindresorhus/gulp-react#usage).  Not sure if `traceur` will cover all our ES6 features, plus it adds additional shims.

2) Just to clarify what is happening with in the `release` step:

1) The `scripts/build.js` takes a template package.json file, pulls the most recent package.json data from the base repo, compiles a new package.json from that template, and sticks it in `build/cjs`
2) the `jsx` command transpiles all the JSX and ES6 into pure Javascript files, but with the same directory structure.
3) I then go the the `build/cjs` directory and hit `npm publish` (after increment the version number before I start the process).

Normally, I prefer Makefiles, and occasionally npm scripts for small jobs. The whole point of this extended `npm scripts` exercise was to do things in a way that would allow Windows users to do development (I was trying to write the commands in a cross-platform manner, but didn't know about `mkdir` with `-p` flag).

But that didn't work out as I planned, so I'm open to trying Gulp.

rrag:
Looking at the `reactify` source, It looks like

reactify just gets the options created for browserify, 

https://github.com/andreypopp/reactify/blob/master/index.js#L9
https://github.com/andreypopp/reactify/blob/master/index.js#L27

so I would just add `"harmony" : true` to the `opts`

That way `watchify` will transform `es6` and `.jsx` on file save

Looking at `gulp-react` plugin source we should be able to figure out a way to get the output of 

`jsx --harmony -x jsx src build/cjs && jsx --harmony src build/cjs` also via gulp, but till that time can we use the `npm run release` ?

It should also be possible increment the version and also do a `npm publish` from a gulp task, but that is after the solution to `gulp release`

I can't get to this gulpfile change till tomorrow morning, do you mind removing `es6ify` and adding the `harmony : true` option and giving it a try?

esbullington:
Thanks @rrag had a busy yesterday.  I'll take a look and try to merge this today, as well as create a new build of our website so that your new candlestick chart is showing up.

rrag:
I faced some issues with making React & d3 external, The usage of browserify-shim had a problem with the multiline strings in docs/examples/main.js

That problem was not there if I reversed the order of transforms (reactify first and browserify-shim next) in package.json but then I was not able to remove React and d3 from the bundle. I finally ended up with browserify-global-shim.

rrag:
@esbullington no worries. 

I don't like browserify-global-shim, because it just replaces `require('react')` with `window.React`, but browserify-shim replaces

`require('react')`
with
`(typeof window !== "undefined" ? window.React : typeof global !== "undefined" ? global.React : null);
`

I am not sure if there is a better way to solve this. The docs and examples I find everywhere is not very helpful for this problem.

frederickfogerty:
Why do you want to make react and d3 external? 

I guess the only reason to do this in my eyes would be if the end user doesn't use browserify/webpack.

rrag:
including external files in a bundle is ok for enterprises and private work, but this is open source and we don't know how react-d3 will be used, it could be used on a website which already has d3 or React or both available. Also they might even have a modified or a smaller version of d3 with just the necessary included using a custom smash build.

I see leaving external out of the build provides flexibility for the user. Looking at examples of other projects (read Backbone) where they use jQuery as a dependency. it is external to the bundle so anyone using it can decide if they want 1.9.x or the 2.x or even a shim like zepto.

Are you saying if the end user uses browserify/webpack and also use React/d3 the uglify/dedupe would remove it hence it is not a problem?

esbullington:
@frederickfogerty yes, browserify/webpack will be the primary way users will use the library, but I've been publishing a minified build to use with JSFiddle examples (mentioned in the README).  And there will be other similar use cases.

I'm glad @rrag figured out how to do an external build because bringing two Reacts into the `window` namespace causes some error messages (@rrag did you happen to test it out yet on JSFiddle?).

It's looking good @rrag I should be able to test and merge this later today.

rrag:
no I did not test in fiddle. If you want me to I can do that tomorrow.

rrag:
also how cool is the watch with browserSync !!! :)

frederickfogerty:
The points you raise are all valid.

An interesting thing I realised was that atm the repo just transpiles the jsx (for npm at least), and leaves it to the user to compile it down with browserify/webpack. I'm not sure this is the ideal method, many other libraries at least use a closure to make nothing is leaked to the global namespace.

I loved using browsersync, but now that I've moved to webpack and react-hot-loader, I can't use it anymore (well, not easily). And to be honest, react-hot-loader is pretty cool too.

rrag:
Deviating off topic here, but could not resist.

> but now that I've moved to webpack and react-hot-loader, I can't use it anymore (well, not easily). And to be honest, react-hot-loader is pretty cool too

True that. I just started to create a webpack with react hot loading, it is very cool. And a huge productivity booster. I am currently getting both browserSync and hot-loader to get to work together so I get the best of both these.

frederickfogerty:
This is where Gitter comes in handy, for chats like these.

I'd be interested to see a working gulpfile/webpack config for this when you're done.

It doesn't quite work with some implementations of Flux (I'm using alt atm), or when you are checking a `componentDidMount()`, but otherwise it works quite well.

esbullington:
@frederickfogerty commonjs specifies implicit local scope so nothing leaks globally unless you use globals.  We're not using globals in this package, so there should be no problem.  

Regarding closures, behind the scenes, that's how commonjs implementations like Browserify work: it's wrapping each module in a closure.

And the reason I don't compile everything together is because I want users to have the option to require a single chart and not have to pull in all the other charts (I'm considering moving the charts each to separate packages as part of a Github react-d3 org, with the component charts in their own library, but that's another topic that I'll open an issue up for).

@rrag I'll try to test it in JSFiddle tonight.

frederickfogerty:
I think keeping them in the same repo (not org) has several benefits: consolidation of issues/PRs, consolidation of interest, reduced effort in contributing (not having to switch directories, etc), less duplication of meta-files (build tools, package.json, etc). The graphs also just aren't big enough to warrant their own repos, imo. 

Changing the method of requiring from `import {ScatterChart} from 'react-d3';` to `import {ScatterChart} from 'react-d3/ScatterChart';` is the best way going forward, to keep the imports light-weight.

frederickfogerty:
Just a thought, maybe @rrag could contribute. If I have files
```js
// myLib.js
exports.a = require('a.js'); // 10kb file
exports.b = require('b.js'); // 100Gb file
```

```js
// myComponent.js
var a = require('./myLib.js').a;
```

does the resulting package end up being ~10kb, or ~100Gb? 

I assume, if this kind of optimisation is not present now, it will be here in the near future. This would solve this bloat/un-used dependency issue.

esbullington:
>I think keeping them in the same repo (not org) has several benefits: consolidation of issues/PRs, consolidation of interest, reduced effort in contributing (not having to switch directories, etc), less duplication of meta-files (build tools, package.json, etc). The graphs also just aren't big enough to warrant their own repos, imo.

These are good points.  I've definitely not made up my mind to move things.

>Changing the method of requiring from import {ScatterChart} from 'react-d3'; to import {ScatterChart} from 'react-d3/ScatterChart'; is the best way going forward, to keep the imports light-weight

Right now (and since I first created the repo), you can `var BarChart = require('react-d3/barchart').BarChart;` without pulling in all the other charts.  I recently added an example of this approach in the README; for more info, see #120  This is the reason I don't compile them all together.

frederickfogerty:
@esbullington I saw your comment about that before, I was just supporting that method of requiring them.

esbullington:
@frederickfogerty Understood, I misread your comment.

@rrag I tried it out, `gulp watch` works great.  Did you find out a way to do emulate the `release` script command with Gulp?  If not, you want to just remove that task and I'll go ahead and merge in anticipation of getting it sorted out soon (I don't really want to carry two different build systems).

rrag:
@frederickfogerty 

> does the resulting package end up being ~10kb, or ~100Gb?

I am not so sure but I guess it would create ~100Gb. since you required the `myLib.js` completely

`var a = require('./myLib.js').a;`

However if you were to `var a = require('./myLib.js/a')`  I am thinking the optimizations you mention are possible. I should try that out sometime.

rrag:
All npm scripts accounted for.

esbullington:
@rrag well-done!  Thank you.  I'm going to merge now..

PR #125, opened 2015-02-18
Brush Support

frederickfogerty:
Hey

We love to use brushes internally, so I thought I'd build it into this library, and see what you guys think. It uses d3's brush utilities, meaning I have to use d3's DOM manipulation - which I'm really not happy about, as it goes against this repo's whole philosophy. I didn't quite want to rewrite a brush implementation in pure React yet, but maybe will in the future.

Unfortunately, it's also not compatible with Voronoi, or any other kind of interaction, as the brush sucks up all pointer events. We've tried to fix this internally, but it pretty much it ends up being hacky.

The changes to `scatterchart.jsx` are only so you can see it working in the examples.

Suggestions/criticisms welcome and appreciated.

### DIFF
--- a/docs/examples/main.js
+++ b/docs/examples/main.js
@@ -178,6 +178,44 @@ var Demos = React.createClass({
         </div>
         <div className="row">
           <div className="col-md-6">
+            <ScatterChart
+              data={scatterData} width={500} height={400} title="Scatter Chart with Brush" brushEnabled={true}/>
+          </div>
+          <div className="col-md-6">
+            <pre ref='block'>
+              <code className='js'>
+              {
+`var scatterData = [
+  {
+    name: "series1",
+    values: [ { x: 0, y: 20 }, ..., { x: 24, y: 10 } ]
+  },
+  ....
+  {
+    name: "series3",
+    values: [ { x: 70, y: 82 }, ..., { x: 76, y: 82 } ]
+  }
+];`
+              }
+              </code>
+            </pre>
+            <pre ref='block'>
+              <code className='html'>
+              {`<ScatterChart
+  data={scatterData}
+  width={500}
+  height={400}
+  title="Scatter Chart"
+/>`}
+              </code>
+            </pre>
+          </div>
+        </div>
+        <div className="row">
+          <hr/>
+        </div>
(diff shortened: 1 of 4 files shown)


frederickfogerty:
Hey @esbullington, I'm not sure if you've seen this yet; in either case, I've just got the brushes working without using d3's brushes, just using mouse events. Let me know what you think.

esbullington:
Oh this is cool, well done.  Glad I didn't close it (since I wasn't going to accept something that manipulated the DOM).

Have you tried resizing the chart based on the brush region?  This gets to a tough question, which is where to store the chart state.  I've been thinking about making a state mixin for this kind of thing.

FRP is ideal for this kind of thing, and we've set up the charts in a functionally-friendly manner.  So that's worth considering, too.

And hmm, trying to think how this could be combined with the Voronoi.  That's a tough one.

frederickfogerty:
I'm glad you like it. I recommend keeping this open until I've solidified an implementation (event publishing, dragging the extent, single-axis/dimension brushing, etc). 

I'm probably going to look at combining this, the Voronoi, and all mouse events (such as scroll, drag, etc) into one `ChartOverlay` component, since you cannot selectively pass through mouse events in current DOM implementations.

Regarding chart resizing, I have a flux/dc-esque implementation which handles this for me. I think a nice way of doing this (visually) would be the way c3 does it (http://c3js.org/samples/options_subchart.html). Technically, I think the data passed via props should always be the overall data set, and then the graph can control its own filter/zoom state, but can also respond to a `filter` prop, which would override the zoom of the graph. This would allow simple and advanced use cases.

frederickfogerty:
The filter/zoom state could be implemented in a mixin (or inherited, when es6 classes arrive).

yang-wei:
This will be really cool if it's working with voronoi
 :+1: @frederickfogerty

rrag:
when you talk about filter can you clarify what you mean

that is just the implementation of `onBrush` ?

esbullington:
Closed since we've moved to a different project structure, and because of the DOM manipulations discussed.  Perhaps I can revisit brush support sometime in the future.

PR #138, opened 2015-02-23
Axis labels added for all applicable charts and in examples

frederickfogerty:


esbullington:
This is much needed. However, I had just started work on this over the weekend, and I was taking into account whether or not the axes were top or bottom, left or right (I've had to do some other refactoring for that to work, as well).  Also, I'd prefer for the end user not to have to worry about setting the labelOffset in the axes, nor to even worry about setting that in the default props in the charts, since it's set to the same values in most of the charts.

So unless you feel like making those changes to this PR, I'm probably going to just merge my axes labels in with the 0.3 branch in the next few days.  If you make the changes, I'll just discard my branch and use yours.

Also, with the possible exception of finishing up this as I describe above (if you want), you may want to hold off on PRs for a few days while I refactor master and publish v0.3.  Chances are I won't be able to use any new PRs since I'll be doing so much refactoring.  But after that, submit away....

And thanks again for all the other ones that I did just merge!

frederickfogerty:
Personally I like giving the user the power to change the offset, that's why I allowed it to be passed down as props. And labeloffset doesn't need to be included, I just made it a prop for the example graphs. 

Maybe I could do some more intelligent defaults for labeloffset. 

I didn't check whether the axis was on the top or bottom, I didn't know the axes could do that, that's awesome. 

I'm not too fussed whether you merge this or not, it was again just something we needed internally.

esbullington:
Gotcha, then it's probably not worth the pain rebasing.  In any case, appreciate you passing on your work.  

And yeah, I'm all for allowing users to set the offset if they need, but I don't want to *make* them pass it.  Like you said, intelligent defaults.

frederickfogerty:
Rebased this to make it easier to merge. How's your axis stuff going @esbullington ?

esbullington:
Closed since we have this functionality now.

PR #139, opened 2015-02-23
Semantic classname

yang-wei:


esbullington:
Hey @yang-wei I'm going to go ahead and merge this, but there's one change I may make in the future, which is to change:

    className="rd3-linechart-axis y axis"

to just:

    className="rd3-linechart y axis"

We shouldn't have to repeat the "axis", since it's already in combined class name.  We should just have the basename `rd3-linechart` and then whatever extra class names we need.   Same goes for the other charts.

Does this sound OK?

yang-wei:
Thanks @esbullington

PR #142, opened 2015-02-27
allowed linechart series to define their own x and y accessors

jakepusateri:
I have data that looks like this: 
```javascript[
[
  {
    month: 0
    value1: 1234,
    value2: 4567
  },
  {
    month: 1
    value1: 7890,
    value2: 7410
  },...
]
```
I would like to be able to not have to preprocess it into separate series in order to graph `value1` vs `value2`. This change allows a `yAccessor` or `xAccessor` to be defined in the series that would override the default selector. I have modified only the LineChart, but I imagine this would impact all of the charts to be done properly. I can modify the others as well if this is the right approach.

### DIFF
--- a/src/linechart.jsx
+++ b/src/linechart.jsx
@@ -386,8 +386,8 @@ var LineChart = exports.LineChart = React.createClass({
             fill={props.colors(idx)}
             pointRadius={props.pointRadius}
             key={series.name}
-            xAccessor={props.xAccessor}
-            yAccessor={props.yAccessor}
+            xAccessor={series.xAccessor || props.xAccessor}
+            yAccessor={series.yAccessor || props.yAccessor}
             interpolationType={interpolationType}
             displayDataPoints={props.displayDataPoints}
           />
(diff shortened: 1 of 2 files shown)


esbullington:
Thanks for the PR. Please note that we have already discussed different data formats several times in the past, so you may want to read those discussions/issues a bit to understand why we've decided on this data format.  This is a very common data format among JS chart libraries (Highcharts, NVD3, Vega chart specification) that I'm familiar with, and the rest of our API depends on having discrete series objects to store things like "series name" and eventually other series-related info.  We've already changed data formats one time, so I'm reluctant to change them again, or to support multiple different formats.

Also, it's not clear to me how your PR relates to the example data you show above.  Your PR changes look for an xAccessor and yAccessor in *each* array object (ie, in each series).  This would allow users to set different x or y accessors for each series.  Is that what you're wanting to do?

Because from the example data above, as well as your description, it looks like you're wanting to support a single array of objects, with each object key referencing a different series.  Supporting this data format would require a lot more changes that what's in this PR, even for just the linechart.

If you'd like to discuss it after you've read through our prior discussions on data formats, please feel free. I'm open to discussion but I'm unlikely to support multiple different formats.

jakepusateri:
After reading the previous discussions it seems that transforming the data during render is a better option.

PR #154, opened 2015-03-07
JSFiddle distribution fix

esbullington:


rrag:
I could not get to it earlier. As promised
Here is a [JSFiddle](http://jsfiddle.net/rrag/8617n0qk/) with the react-d3.js created from `gulp minified`.

PR #157, opened 2015-03-11
Improve code (following guideline and use es6 features that are supported)

yang-wei:
Browse through all code base because din't catch up with so many changes. So just did this.

### DIFF
--- a/src/areachart/AreaChart.jsx
+++ b/src/areachart/AreaChart.jsx
@@ -81,7 +81,7 @@ module.exports = React.createClass({
 
     var layers = stack(props.data);
 
-    var trans = "translate(" + props.margins.left + "," + props.margins.top + ")";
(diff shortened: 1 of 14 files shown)


esbullington:
Hey, this looks great!  I'm going to merge this now as is, but one favor to ask for future commits: I'm trying to do a better job of squashing my commits down to the smallest possible number before submitting a PR.  In future PRs, could you please squash your commits into the smallest possible number?  I mean, don't squash unrelated commits, but if you have several commits that deal with 1 topic (like es6 object shorthand), then it would be great if you could squash those.

Also,  I usually use single-quotes in JS and double-quotes for HTML and JSX attributes.  However, after thinking about it, I think it will be ok for us to use single-quotes in JSX.  So merged -- thanks again!

yang-wei:
++ for smallest possible number of commits @esbullington

PR #158, opened 2015-03-14
Fixed syntax error in piechart/arc.jsx that broke the build

grahamscott:
Hi

I noticed this small syntax error that stopped gulp from doing it's thing, and also broke the tests.

Cheers
Graham


yang-wei:
good catch @grahamscott . Thx !

PR #159, opened 2015-03-15
Moving utils directory into tests directory

yang-wei:


esbullington:
Hey @yang-wei thanks!  Traditionally, `utils/datagen.js` has also been used in  `docs/examples/main.js`, so that path is incorrect now.  However, we no longer even use the `datagen.js` module there anymore, so if you just delete that require in the `docs/examples/main.js` file, everything should work fine.

I look forward to merging it as soon as you do this.

PR #166, opened 2015-03-25
Add browserify transform field to package.json

bjoerge:
Browserifying a project that depends on react-d3 from a local checkout (via `npm link`) currently fails with parse errors due to the jsx/es6 syntax in source files. This PR adds a [transform](https://github.com/substack/browserify-handbook#browserifytransform-field) field to package.json, which makes browserify transpile the source code on the fly in cases like this.

This is super useful for debugging purposes and should cause no additional overhead when react-d3 is installed from npm, as a different package.json is shipped with the npm release.


esbullington:
Thanks!

PR #168, opened 2015-03-25
Conforms area chart to use index instead of name for selecting color

evenwestvang:
Use index instead of name in line with convention so that the legend corresponds to the chart.

### DIFF
--- a/src/areachart/DataSeries.jsx
+++ b/src/areachart/DataSeries.jsx
@@ -18,7 +18,7 @@ module.exports = React.createClass({
     var path = area(props.data);
 
     return (
-      <Area fill={props.colors(props.name)} path={path} />
+      <Area fill={props.colors(props.index)} path={path} />
     );
   }
 


evenwestvang:
Passes locally. Mysteriously failing in CI on setup.

esbullington:
Thanks for your PR.  The test is failing because of some issue with the chai dependencies, not your code.  I'll sort it out when I get a block of free time to work with this, probably tomorrow or next day.

esbullington:
Hey, would you mind merging or rebasing (whichever you prefer) your branch against master and then pushing the changes?  It should take care of this build error.  Thanks!

evenwestvang:
Rebased against master and updated. Sorry about the delay. Will be more prompt in the future.

esbullington:
No problem, there's no rush.  For some reason, it still doesn't look like you got the latest commits in your branch from the upstream master, one of which contains an update to the package.json that resolves the Travis error (which has nothing to do with your code).   Are you sure you [rebased against upstream master](https://robots.thoughtbot.com/keeping-a-github-fork-updated)?

Thanks for your patience and sorry about the dependency problem -- we'll get it figured out.

evenwestvang:
That's weird - the [commit history](https://github.com/bengler/react-d3/commits/area-chart-color-fix) it has the 4 upstream patches from 17 hours ago as well as Bjørge's patch from yesterday. Does it look right to you?

evenwestvang:
On the other hand, Github was sideways from the Baidu DDOS when I pushed it and it only went through on the fourth try. Perhaps manually pull the Travis lever?

esbullington:
Thanks, that did it.  Not sure if I was looking at the wrong branch on your repo (still drinking my coffee, so it's a possibility), or if the DDoS delayed things showing up because of some caching issue (I had to hit refresh several times, and it was right after you pushed your changes).

In any case, merged and thanks for your help, much appreciated!

PR #170, opened 2015-03-27
added tickStroke for AxisTicks and cleaned up AxisTicks

jakepusateri:
added propTypes and tickStroke to AxisTicks
consolidated xScale and yScale into scale prop
corrected tick placement for all orients


yang-wei:
thanks @jakepusateri

PR #187, opened 2015-04-18
Key standardization and LineChart key bug fix

MatthewHerbst:
This pull request addressed two things.

1. LineChart was passing only the `series.name` as the key to each `DataSeries` rather than passing `series.name + idx`. This was causing possible duplicate key issues and has been fixed.
2. The value of `key` throughout the code was inconsistently generated (sometimes `series.name`, sometimes `idx`, and sometimes `series.name + idx`). Where possible, the `key` value has been standardized to `series.name + idx`

I also added two PropTypes checks, though many, many more are missing. I will add them in when I get a chance.


MatthewHerbst:
Not sure why the build failed - I double checked the code and it doesn't seem there are any changes that would cause a test failure. Is there any way to makes the tests more verbose?

esbullington:
Thanks for your PR.   I agree that we've used several different ways of generating keys and it would be nice to standardize whenever possible.  I was planning to standardize to just `idx`, which is guaranteed to be unique within the collection of child components (which is all the key needs to be).  The reason we need `seriesName + '-' + idx` in the `id` property is because we need those id's to be globally unique for the voronoi overlay (and we have to use a different prop other than `key` because React makes the key prop an essentially [private property](http://facebook.github.io/react/blog/2014/10/16/react-v0.12-rc1.html#breaking-change-key-and-ref-removed-from-this.props)).

Your bar chart test failed because there are no default labels set yet you're depending on those labels to be present for the key.  Since they're undefined, when you concatenate that with an index, you get `NaN`, which is presumably not a valid key (and so the component fails to render).  The quick fix for that is just to use the index for the key, as I describe above.

Minor suggestion: could you leave the `seriesName` variable name as it is now in the child components?  I'd prefer to encourage more descriptive variable names whenever possible (in the parent component, it's clear what `name` refers to since it's a property of the `series` object).

And thanks for adding the missing prop types, I'm sure we're missing quite a few...

MatthewHerbst:
Ah, that makes sense regarding the BarChart test, thanks for the explanation. And sure, happy to use `seriesName` and just `idx`. I'll update soon.

yang-wei:
lgtm

Issue #188, opened 2015-04-18
bar chart with negative numbers


kriswallsmith:
I need to be able to render a chart with negative numbers, like so…

![screen shot 2015-04-18 at 8 35 49 am](https://cloud.githubusercontent.com/assets/33886/7216069/275d7b9c-e5a6-11e4-9d77-34f5af68fc0c.png)

Unfortunately a data array with a negative value throws an error…

>Error: Invalid negative value for <rect> attribute height="-128.46135247235017"

jsanchez034:
Hey @kriswallsmith, @esbullington, have a fix for this issue and I'm about create the pull request for the fix

jsanchez034:
Hey @esbullington here is my pull request with fix, #189.  Let me know what you think.

yang-wei:
@kriswallsmith Give it a try on latest master =)

kriswallsmith:
That works. Thank you!

PR #189, opened 2015-04-18
Fix for exception thrown and rendering problems for BarCharts with negative values #188

jsanchez034:


jsanchez034:
Hi @yang-wei, no worries. I just made the variable name change let me know if anything else needs changing.

yang-wei:
Thanks @jsanchez034 .

PR #190, opened 2015-04-19
Pie chart formatter

### LINKED ISSUE #178: Unable to turn off % at the end of a value in a PieChart.
Always displays a percentage even when not desired.  Are there props that could be exposed to turn off or switch the suffix?

yang-wei:
This fix #178. Now Pie Chart value can be formatted

`valueTextFormatter` is default to `{val=>val+'%'}`

```xml
// eg: To add a dollar sign in front of value
<PieChart data={} valueTextFormatter={val=>'$'+val} />

eg: To remove %
<PieChart data={} valueTextFormatter={val=>val} />
```


dennisjac:
I just tried the dollar example "valueTextFormatter={val=>'$'+val}" for my PieChart but this doesn't change anything. Is this actually working?

PR #193, opened 2015-04-21
Adds showInnerLabel and showOuterLabel props to PieChart

samdalton:
I want to have a custom legend for the PieChart, and don't want to show inner labels for a small chart (< 50px) where the text labels will not fit. The best option I saw was to allow disabling of the inner and outer labels via props to PieChart: `showInnerLabels` and `showOuterLabels`. Let me know if you had other thoughts on this or it doesn't look right.

### DIFF
--- a/src/piechart/Arc.jsx
+++ b/src/piechart/Arc.jsx
@@ -16,13 +16,17 @@ module.exports = React.createClass({
     innerRadius: React.PropTypes.number,
     outerRadius: React.PropTypes.number,
     labelTextFill: React.PropTypes.string,
-    valueTextFill: React.PropTypes.string
+    valueTextFill: React.PropTypes.string,
+    showInnerLabels: React.PropTypes.bool,
+    showOuterLabels: React.PropTypes.bool
   },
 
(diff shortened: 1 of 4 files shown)


src/piechart/Arc.jsx
```diff
@@ -45,12 +49,9 @@ module.exports = React.createClass({
...
 
-    return (
-      <g className='rd3-piechart-arc' >
-        <path
-          d={arc()}
-          fill={props.fill}
-        />
+    var outerLabels = null;
+    if (this.props.showOuterLabels) {
+      outerLabels = [
```
yang-wei: Now `outerLabels` is an array. 
This should be `(` instead of `[` ?

samdalton: I wasn't able to have to adjacent JSX elements assigned to this var (syntax error), so they're in an array to be passed in below.

samdalton:
@yang-wei squashed!

yang-wei:
Thanks @samdalton ! Never know jsx can be wrapped like that.

PR #194, opened 2015-04-21
Added stroke support to PieChart to enable lines between

samdalton:
Passing in a `stroke` prop that is a string color (e.g. "white") to PieChart will render a white border around each segment, producing a chart like:
screen shot 2015-04-21 at 4 20 24 pm

### DIFF
--- a/docs/examples/main.js
+++ b/docs/examples/main.js
@@ -315,7 +315,7 @@ var Demos = React.createClass({
         </div>
         <div className="row">
           <div className="col-md-6">
-            <PieChart data={pieData} width={450} height={400} radius={110} innerRadius={20} title="Pie Chart" />
+            <PieChart data={pieData} width={450} height={400} radius={110} innerRadius={20} sectorBorderColor="white" title="Pie Chart" />
           </div>
           <div className="col-md-6">
             <pre ref='block'>
@@ -338,6 +338,7 @@ var Demos = React.createClass({
   height={400}
   radius={100}
   innerRadius={20}
+  sectorBorderColor="white"
   title="Pie Chart"
 />`
                 }
--- a/src/piechart/Arc.jsx
+++ b/src/piechart/Arc.jsx
@@ -17,6 +17,7 @@ module.exports = React.createClass({
     outerRadius: React.PropTypes.number,
     labelTextFill: React.PropTypes.string,
     valueTextFill: React.PropTypes.string,
+    sectorBorderColor: React.PropTypes.string,
     showInnerLabels: React.PropTypes.bool,
     showOuterLabels: React.PropTypes.bool
   },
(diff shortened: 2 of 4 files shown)


danpatnate:
:+1:

yang-wei:
Thanks ! This is probably needed. 
But could you name it more specifically instead of just `stroke` ?
```
// not so clear
<PieChart stroke={...} />
// clearer
<PieChart segmentBorder={...} />
```

esbullington:
This looks very nice, thanks!  I agree with @yang-wei that a self-documenting property name would be much appreciated here.  We've been sometimes using the svg attribute and property names for the lower-level chart components, but for the property names of components that are part of the library's public API, I'd prefer to use descriptive names (this will help us with the documentation).

But the borders definitely make the chart easier to read.  And the white borders go very nicely with that palette.  Are you using `d3.scale.category10()`?

samdalton:
@esbullington @yang-wei Thanks for the feedback, I was a little unsure on this, too. Changed it to `segmentBorderColor` and added example usage to the doc page. You're right, it is much clearer now.

The colors are from our company color palette, actually.

parshap:
I don't mean to bikeshed, but may I suggest `arcStroke` to stay consistent with existing nomenclature in react-d3 (Arc) and in svg (stroke)? `arcStrokeWidth` and a few of the other `stroke-` properties may be useful to add later too.

yang-wei:
Exactly @parshap . Although `segmentBorderColor` sounds clear for me but being semantic is the king. What do you think @samdalton ?

esbullington:
Thanks, @samdalton   `segmentBorderColor` is just fine for now.  I think we're going to need to open an separate issue for naming and establish some clear written guidelines for naming to put in the `CONTRIBUTING.md` file (yang-wei already has written some guidelines for CSS naming, but we've not yet generalized it).

The problem with the name `arcStroke` is that `arc` refers to just the [this part](http://www.google.com/imgres?imgurl=http://www.mathwords.com/a/a_assets/arc.gif&imgrefurl=http://www.mathwords.com/a/arc_circle.htm&h=146&w=180&tbnid=fdxY0F5crETWFM:&zoom=1&tbnh=116&tbnw=144&usg=___q_u_vby_uDzQiQvXA0a-h0TkOA=&docid=VCkdSu_U4HpHzM) of the circle, whereas samdalton's PR puts a border around the entire pie slice (since the whole slice is a single `path`, putting a border just on the arc would be tricky).  Technically (and upon confirming with Wikipedia), the pie slice would be the "sector", but I think segment is probably OK for now.

Anyway, I'd like to go ahead and merge this, but it needs to be rebased.  If you don't get a chance to do that today, I'll rebase and merge this evening.  Thanks again!

samdalton:
@esbullington @parshap I went with the maximally semantic `sectorBorderColor`, and rebased master so this should be good to go.

parshap:
:+1:

PR #195, opened 2015-04-22
Pass tickValues prop to axes

parshap:
Add `xAxisTickValues` and `yAxisTickValues` props to charts which will get passed down to `XAxis` and `YAxis` components as the `tickValues` prop.

Only one of `*AxisTickValues`, `*AxisTickInterval`, `*AxisTickCount` can be used at a time. This should probably be documented somewhere, but I wasn't sure where. Maybe a custom propTypes validator?


esbullington:
Thanks!  I've been meaning to do this.

PR #196, opened 2015-04-22
Add interpolation to area chart

parshap:
I've implemented the the same interface to control interpolation as currently implemented in `LineChart` (using the `interpolate` and `interpolationType` props).

I also added some missing propTypes to `LineChart`.


esbullington:
Thank you!

PR #197, opened 2015-04-22
Optional chart titles

parshap:
I need charts to render without a title and currently even if you don't pass in a `title` prop, an `h4` element is rendered. This changes `BasicChart` and `LegendChart` to not render any unnecessary elements.

### DIFF
--- a/src/common/charts/BasicChart.jsx
+++ b/src/common/charts/BasicChart.jsx
@@ -6,16 +6,43 @@ module.exports = React.createClass({
 
   displayName: 'BasicChart',
 
-  render: function() {
-    return (
-      <div>
+  propTypes: {
+    title: React.PropTypes.node,
+    viewBox: React.PropTypes.string,
+    width: React.PropTypes.number,
+    height: React.PropTypes.number,
+    children: React.PropTypes.node,
+  },
+
+  _renderTitle() {
+    if (this.props.title != null) {
+      return (
         <h4>{this.props.title}</h4>
-        <svg
-          viewBox={this.props.viewBox}
-          width={this.props.width}
-          height={this.props.height}
-        >{this.props.children}</svg>
-      </div>
+      );
+    }
+  },
+
+  _renderChart: function() {
+    return (
+      <svg
+        viewBox={this.props.viewBox}
+        width={this.props.width}
+        height={this.props.height}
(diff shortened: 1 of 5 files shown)


yang-wei:
Thanks for committing =) And also the test.
Just one thing, is there any reason not to return `null` or nothing on `_renderTitle()` method.
```js
  _renderTitle() {
    if (this.props.title != null) {
      return (
        <h4>{this.props.title}</h4>
      );
    } /* added this*/ else {
      return null;
    }
  },
```
So that we can do this:
```js
render: function() {
   // if (this.props.title != null) {
      return (
        <div>
          {this._renderTitle()}
          {this._renderChart()}
        </div>
      );
   /* }
    else {
      return this._renderChart();
    }
  }*/
```

esbullington:
Thanks for the PR and particularly the tests, @parshap !

And interesting point, @yang-wei I just read [this React blog update on returning null.](https://facebook.github.io/react/blog/2014/07/17/react-v0.11.html#rendering-to-null)  I didn't know we could return `null` in JSX now (as of React 0.11) -- I've still been returning `<g/>` or `<span/>`.  The `null` currently renders as a `<noscript/>` (and so no effect on layout) but future versions may not render any tag.

Anyway, it's a good point, but I'll leave it up to @parshap and will accept the PR either way.  One thing: if parshap does elect to make this change to something like yang-wei proposes, I'd prefer to omit the `else` statement altogether whenever the `if` statement returns, since that's consistent with the rest of our code base (hopefully!).  So, preferred would be:

```js
_renderTitle() {
  if (this.props.title != null) {
    return <h4>{this.props.title}</h4>;
  }
  return null;
}
```

I know different people and projects have different opinions about this stylistic issue, but I'd prefer to stick with the above approach for this project.  As you can see, I also tend to leave off the parentheses when returning short JSX snippets, but that's up to the individual contributor.

parshap:
No problem! Thanks for the great project.

> Thanks for committing =) And also the test.
> Just one thing, is there any reason not to return null or nothing on _renderTitle() method.
> 
> So that we can do this:

I changed `_renderChart()` to return `null`, but I'm trying to avoid the extra `<div>` wrapper, so I've left the branching in `render()`.

Also rebased against latest master.

PR #198, opened 2015-04-22
Bug fix in linechart/DataSeries

MatthewHerbst:
Fixed a bug in the linechart/DataSeries component where the render method was polling the first value of the data array to check for the x-axis data type, without first checking to see if the element of the data array was defined. A check has been placed into DataSeries's render() to check the length of the array first, returning null if it is empty. Note that this check does not cover sparse arrays, which will still potentially cause the render to fail.

I also added a PropType for seriesName in DataSeries and added a default value for it since it is required by portions of the code, such as creating an id. I also removed the PropType for data from linechart/Line since the data prop is not used by Line, and, it was the wrong type anyways.

### DIFF
--- a/src/linechart/DataSeries.jsx
+++ b/src/linechart/DataSeries.jsx
@@ -12,6 +12,7 @@ module.exports = React.createClass({
 
   propTypes: {
     data: React.PropTypes.array,
+    seriesName: React.PropTypes.string,
     interpolationType: React.PropTypes.string,
     fill: React.PropTypes.string,
     xAccessor: React.PropTypes.func,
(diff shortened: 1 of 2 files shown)


MatthewHerbst:
I also checked the other DataSeries components and none of them seem to suffer from this problem (in part because LineChart uses its DataSeries in a different way than the others - something that should be made consistent eventually).

yang-wei:
Hi @MatthewHerbst , it would be great if you can squash the `Fixed typo` commit ?

MatthewHerbst:
Done - didn't know I could do that - thanks for the learning!

PR #203, opened 2015-04-30
New feature: props.colorAccessor

MatthewHerbst:
**tl;dr:** this is a non-breaking change that optionally gives chart users more control over how colors are generated from data passed into the chart.

Added a new feature, props.colorAccessor. props.colorAccessor is a callback function that takes two arguments, an element and an index. It is meant to be called inside any call to props.colors. 

For example, a call that was previously 

```javascript
props.colors(idx)
``` 

is now 

```javascript
props.colors(props.colorAccessor(d, idx))
``` 

This allows chart users to base colors on more than just an index in an array. Further, this allows the user to not have to retain a copy of the data provided to the chart. 

For example, in the current way of doing things, if I want to base colors off of anything other than the index, I need to do something along the lines of:

```javascript
var MyCoolComponent = React.createClass({
  getDefaultProps: function() {
    return {
      colorScale: d3.scale.category20()
    };
  },

  colors: function(idx) {
    var color = this.props.colorScale(props.myData[idx].uuid);
    return color;
  },

  render: function() {
    <LineChart 
      ...
      colors={this.colors}
      ...
    />
  }
});
```

with props.colorAccessor, I can simply do:

```javascript
var MyCoolComponent = React.createClass({
  getDefaultProps: function() {
    return {
      colorAccessor: (d, idx) => d.uuid
    };
  }

  render: function() {
    <LineChart 
      ...
      colorAccessor={this.props.colorAccessor}
      ...
    />
  }
});
```

The props.colorAccessor feature has been added to all charts except for BarChart. This is because BarChart does not currently support multi-colored bars, instead using the same fill for each bar. I will add support for multi-colored bars soon which will also include the props.colorAccessor feature.

For most of the charts the needed changes to support the new feature were fairly minor. However, some charts needed a bit more changes to support the feature.

Along with the above feature, code styling and formatting has been applied where appropriate to the various touched files. This includes modification of PropTypes, removing unneeded calls to `this`, as well as some (imho) much needed spacing.

Also included in this branch is the addition of .DS_Store to .gitignore, and the changing of the link in CONTRIBUTING.md to the Google JavaScript Style Guide from HTTP to HTTPS.


esbullington:
Thanks for the PR.  Many of these look like good ideas/changes, but they've resulted in a lot of changes to the code base, so I'm going to need to take a closer look over the weekend when I have more time.

@yang-wei do you think you might take a look if/when you get a chance to make sure there are no breaking changes?  A lot of the changes were in the charts you've been working with.

yang-wei:
@MatthewHerbst Looks good to me. 


Just one thing:
Since the first argument of `colorAccessor` will always be element, so it will be good to have it named `element`(or something like this) instead of just `d` in `getDefaultProps`.

```js
// instead of this
colorAccessor: (d, idx) => idx 
// better ?
colorAccessor: (ele, idx) => idx
```

Also this is really a good feature but this would be hard to understand for those who are not familiar with d3. It would be really great if a test regarding this feature is added to any chart.

Cheers.


@esbullington One thing here. This pr will change a lot of our code base, especially the indentation on `proptypes`. To me it looks good, just want to make sure that you're ok with them.

MatthewHerbst:
@yang-wei the standard in d3 is to use `d` for `datum`. I'm happy to change it though if you and @esbullington like `ele` (or some derivative) better.

And good point regarding testing! I'll add some in tomorrow. However, I don't think the method can ever fail. For example, the following is valid:

```javascript 
var scale = d3.scale.category20c()
var value = "undefined" // This would be if you didn't pass enough arguments to colorAccessor
scale(value); // "#3182bd"
``` 

So is

```javascript 
var scale = d3.scale.category20c()
var value = {} // This would be if you passed a non-primitive as the second argument to colorAccessor
scale(value); // "#3182bd"
``` 

`colorAccessor` is really just a getter method when you think about it - you usually don't test those much since they are too simple.

The tests I'll add will insure that the proper index is returned when the default colorAccessor is called. Beyond that, the possible return values if a user override the method are too arbitrary (and all valid) to test.

yang-wei:
> the standard in d3 is to use d for datum

This is true. But in this case `ele` seems more appropriate to me. 

> The tests I'll add will insure that the proper index is returned when the default colorAccessor is called. Beyond that, the possible return values if a user override the method are too arbitrary (and all valid) to test.

@esbullington  This will be enough =)  Test will help developers to see how to use the feature.

esbullington:
@yang-wei  thanks for taking a look.  I wouldn't have thought to add the indented `propTypes`, but the PR contained a lot of other good changes, so it wasn't a big deal to me.  Most of all I was concerned about breaking changes since these changes will go out on my next patch release (which I'll do this week).

Let's stick with `d` unless you've already made the changes, in which case `ele` is fine, too.  In general, I've tried to stick as much as possible with d3's stylistic preferences (although I know I've not always done a good job with that).

Any tests for documentation would be welcome, particularly since adding a getter to what is already a somewhat confusing and definitely undocumented feature (the `colors` property) will make it even more confusing to users (hopefully I can combine the two properties when we next batch together a lot of breaking changes).  Even better (and possibly less time consuming) would be to add a brief note to our meager [documentation](https://github.com/esbullington/react-d3/wiki) explaining how to use the feature.  But there are good changes in this PR, so don't want to discourage if you don't have time.

And it'll be necessary to rebase before I can merge.  If you don't have time, I'll do it later when I merge it.

MatthewHerbst:
@esbullington I'll do the rebase and add the tests later this morning. Do you want me to remove the PropTypes indentation? I'll edit the Wiki later tonight.

Also, very excited to see the patch come out! I've been manually making changes to the node_modules/react-d3 that my project is pulling from. I'll make sure this is done soon so it can be included!

esbullington:
@MatthewHerbst great, I'll merge as soon as you get it in (no rush).  And no, the `PropTypes` indentation is fine, I think it probably makes things more readable.  It's just not something I generally do.

One thing that will help us review contributions quicker in the future is to keep PRs as small and as topical as possible.  It's *much* easier for me to review 4 PRs from someone, with each concerning a different issue or topic (say, "indentation", or "adding getter for colors prop"), rather than 1 PR with all the changes combined together.  I know it's not always easy or possible to break things up into small PRs when you're doing general refactoring like you've done here, but whenever possible it's greatly appreciated.

MatthewHerbst:
Sure thing @esbullington I was king of just making those formatting changes to make it easier for me to make the changes I was trying to make. I'll try and keep the two separate next time. Rebase is done - writing the tests now.

MatthewHerbst:
@yang-wei I'm actually starting to think that implementing these tests is not possible without extensive mocking - there is no way for me to access the values being passed into the colorAccessor function. It seems much simpler to edit the Wiki for it, which I will do. We should maybe open a new issue to discuss how we want to test property methods, especially ones which can be passed in by the user.

MatthewHerbst:
@esbullington I've started modifying the documentation. Before I go crazy, is this format ok? https://github.com/esbullington/react-d3/wiki/CartesianChartPropsMixin

esbullington:
@MatthewHerbst looks great!  Documentation in this case will be better than tests, and it looks great.  Maybe rename the page name to `Common Props` or `Shared Props` to make it clearer, and I'll probably rename the mixin file to match, as well.

Do you happen to remember what you ran to rebase?  For some reason, there are two copies of your commits now (each set with different commit hashes), and also a bunch of prior commits are showing up above, even though those don't appear to be copied.  I can fix the history easily, just trying to figure out what exactly caused that to happen.

MatthewHerbst:
Yeah, I noticed that but wasn't sure what it was about.

```git
// From branch matthewherbst-feature/color-accessor, I did:

git fetch upstream
git checkout master
git merge upstream/master
git checkout matthewherbst-feature/color-accessor
git rebase master
// Had to resolve a merge conflict at this point
git commit [the merge conflicts]
git rebase --continue
git push origin matthewherbst-feature/color-accessor
// Then, since I had forgotten to run the tests locally, the build failed
// There was some warning about cherry-picking at this point, so I used the message advice to use --allow-empty
git commit --allow-empty [the fix - removing structure] matthewherbst-feature/color-accessor
git push origin matthewherbst-feature/color-accessor
```

esbullington:
Hey @MatthewHerbst   no problem, I can straighten it out.  I should have some spare time tomorrow morning for this, so I'll edit the history and merge it then.   Promise I haven't forgotten.

Thanks again for the PR and for your patience!

MatthewHerbst:
@esbullington :+1:

esbullington:
Moved to #210

PR #206, opened 2015-05-06
Fix arc key warning

yang-wei:
Quick fix for react key warning and run some lint.


yang-wei:
I added 2 new methods which are `renderOutputLabel` and `renderInputLabel` to decrease the code in render method. In `renderOutputLabel`, warning was given because we return an array 
```js
return [ <text />, <line />, ... ]
```
This pr wrap the returned element with a <g> element.

PR #207, opened 2015-05-06
Single series chart animation 86

yang-wei:


yang-wei:
Haven't add the option to turn off animation

esbullington:
Thanks for remembering the animation toggle flag, looks great :)

PR #208, opened 2015-05-06
Scatterchart series color

yang-wei:
This was my mistake. In scatter chart, all circles are rendered with same color. Color should differ if they are not from the same series. 

Added another property `seriesName` to the objects of allValues in `utils/index.js`. Added a test for this to prevent future mistake.

Also I realised that in those graphs that we are using d3 color palettes, users don't really have a option to color their graph. They can only opt for another 3 color palettes that d3 provided. Might open another issue regarding this if you want a further discussion.

### DIFF
--- a/src/scatterchart/DataSeries.jsx
+++ b/src/scatterchart/DataSeries.jsx
@@ -30,7 +30,7 @@ module.exports = React.createClass({
     var yScale = props.yScale;
     var xAccessor = props.xAccessor,
         yAccessor = props.yAccessor,
-        cx, cy;
+        cx, cy, circleFill;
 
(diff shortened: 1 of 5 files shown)


MatthewHerbst:
If you look at my new feature that I'm merging in #203, it allow the user the full option of giving the color palette and deciding how colors are picked from that palette (via the `props.colorAccessor` method).

Once that

yang-wei:
:+1: @MatthewHerbst 

Ya I almost forgot your handy `color.accessor` feature. Because that involves a lot of changes, so review process slow down. Give us another day, either @esbullington or me will merge in.

PR #210, opened 2015-05-07
Fix rebase of #203: New feature: props.colorAccessor 

esbullington:


MatthewHerbst:
:+1:

esbullington:
Thanks, I was just asking if you in the other PR if you wanted to take a look.  Which chart's `DataSeries` got missed?

I've got to run now but I'll make that change tonight or tomorrow morning.

MatthewHerbst:
@esbullington the `DataSeries` call in `ScatterChart` is the one missing `colorAccessor`

yang-wei:
Looks good to me.

esbullington:
Yeah, the problem is that @yang-wei change from #208 (which was a nice move forward in freeing us from the `immstruct` dependency) means that we have to select the element by serial, otherwise it doesn't work.  I'm going to go ahead and merge it so we can get your changes in.  If you can find a workaround, we'll merge that in before I publish the next release, does that sound good?  Otherwise, we'll just release it as a 0.x release (which is fine).

I won't be able to do any work with this project for the next few days at least because of job duties.  

@yang-wei if you're around, please feel free to merge PRs in the `src/` directory even without a review from me if you feel comfortable with the changes, and I'll take a look at the merged changes whenever I get a chance.

MatthewHerbst:
Sounds good!

Issue #204, opened 2015-05-05
Better way to scale


jmuel:
I was running into an issue getting the scaling to work easily so I made a few small changes shown [here](https://github.com/jmuel/react-d3/commit/9a25085510f390757c759609068ffcbfaee0f057). This isn't a complete solution so I'd have to go through the use cases that would be affected by this, but it works for my use case and can be generalized.

esbullington:
It's a good approach if your parent element scales (like if you're using Bootstrap or some other grid-based CSS system).

You know what, we could just make width and height strings and then users could either put the pixel width/height or `100%`.  That would allow users who have a scalable parent element to easily scale, like you are doing.  Would there be any downside to this approach?

It's a breaking change, so we'd probably want to merge it to a `v0.4.0` branch.

jmuel:
That's basically what I was intending to do. The important thing here is that we have the overall height and width which goes on the svg element itself and should be a string, and we have the internal height that is used for measurements and should be a number. I was thinking we could actually create a separate prop for the first concept and keep the height and width the same, and default the svg's height and width to the internal height and width if there's no scaled values. This would allow us to make it a non-breaking change, but it might have additional complications I'm not thinking about.

esbullington:
Actually, it doesn't even have to be a breaking change even without a separate height.  We could either not validate those propTypes or (better yet) use the React `oneOf` validator:

```javascript
    // An object that could be one of many types
    optionalUnion: React.PropTypes.oneOfType([
      React.PropTypes.string,
      React.PropTypes.number,
      React.PropTypes.instanceOf(Message)
    ]),
```

I've been working in statically-typed languages recently, so I momentarily forgot the fact that I'm working here in JS (`propTypes` trick me into thinking I'm working in static types, even if the validation is in fact runtime).

That said, it's not a bad idea to set the default height and width to `100%`, which should accomplish pretty much what you're suggesting.

yang-wei:
Yes this is not a breaking change. PR are welcome @jmuel

samdalton:
I ran in to this issue as well, but because of AreaChart, you can't accept both a string and a number as it does math with the values. See my comment here: https://github.com/esbullington/react-d3/issues/179#issuecomment-95329569
This would work for LineChart but if we wanted the interface to be consistent across all charts it needs a different approach.

esbullington:
[...]

PR #213, opened 2015-05-14
Added a ViewBoxMixin that lets charts configure their viewbox

samdalton:
Following discussion on this issue here: https://github.com/esbullington/react-d3/issues/204 I implemented a proposed solution and show it working for the AreaChart. The remaining charts have yet to be updated. By using the `ViewBoxMixin` charts can accept a `viewBoxObject` prop that will let charts access outer chart dimensions as pixel values, and generate a string viewBox value that can be set on the svg element. See the AreaChart for an example. Comments/thoughts appreciated - this doesn't need to be merged in but is rather another starting point for the discussions that seemed to have stagnated a bit.

### DIFF
--- a/src/common/charts/LegendChart.jsx
+++ b/src/common/charts/LegendChart.jsx
@@ -11,12 +11,12 @@ module.exports = React.createClass({
     colors:         React.PropTypes.func,
     colorAccessor:  React.PropTypes.func,
     title:          React.PropTypes.node,
-    viewBox:        React.PropTypes.string,
-    width:          React.PropTypes.number,
-    height:         React.PropTypes.number,
+    width:          React.PropTypes.node,
+    height:         React.PropTypes.node,
     children:       React.PropTypes.node,
     legend:         React.PropTypes.bool,
     legendPosition: React.PropTypes.string,
+    viewBox:        React.PropTypes.string,
     sideOffset:     React.PropTypes.number,
     margins:        React.PropTypes.object,
     data:           React.PropTypes.oneOfType([
@@ -47,9 +47,8 @@ module.exports = React.createClass({
           colors={props.colors}
           colorAccessor={props.colorAccessor}
           data={props.data}
-          width={props.width}
+          width={props.sideOffset}
           height={props.height}
-          sideOffset={props.sideOffset}
         />
       );
     }
@@ -69,8 +68,14 @@ module.exports = React.createClass({
     return (
       <div style={{'width': props.width, 'height': props.height}} >
         {this._renderTitle()}
-        {this._renderLegend()}
-        <svg viewBox={props.viewBox} width={props.width - props.sideOffset} height={props.height}>{props.children}</svg>
+        <div style={{ display: 'table', width: '100%', height: '100%' }}>
+          <div style={{ display: 'table-cell' }}>
+            <svg viewBox={props.viewBox} width="100%" height="100%">{props.children}</svg>
+          </div>
+          <div style={{ display: 'table-cell', width: props.sideOffset, 'vertical-align': 'top' }}>
+            {this._renderLegend()}
+          </div>
+        </div>
       </div>
     );
   }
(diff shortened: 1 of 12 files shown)


src/mixins/CartesianChartPropsMixin.js
```diff
@@ -27,12 +27,11 @@ module.exports =  {
     xAxisFormatter:    React.PropTypes.func,
     legend:            React.PropTypes.bool,
     legendOffset:      React.PropTypes.number,
-    width:             React.PropTypes.number,
-    height:            React.PropTypes.number,
+    width:             React.PropTypes.node,
+    height:            React.PropTypes.node,
```
parshap: I think you want `React.PropTypes.oneOfType([React.PropTypes.number, React.PropTypes.string])` here.

src/common/charts/LegendChart.jsx
```diff
@@ -70,7 +72,7 @@ module.exports = React.createClass({
       <div style={{'width': props.width, 'height': props.height}} >
         {this._renderTitle()}
         {this._renderLegend()}
-        <svg viewBox={props.viewBox} width={props.width - props.sideOffset} height={props.height}>{props.children}</svg>
+        <svg viewBox={this.getViewBox()} width={this.getOuterDimensions().width - props.sideOffset} height={props.height}>{props.children}</svg>
```
parshap: I think `width={this.getOuterDimensions().width}` will result in having a static `px` width set on the svg element instead of a `%` that gets passed in. I see this is a complication because of the arithmetic that needs to be done to account for the legend's size. I think the layout system needs to be changed to use some sort of table layout or floats so that the svg can be `width=100%` while still allowing for room for the legend.

esbullington: There is [this approach](http://blog.scottlogic.com/2015/02/02/svg-layout-flexbox.html) using Facebook's css-layout library to implement/emulate a flexbox layout.  However, this comes with a heavy cost when used with React: the background calculations for SVG could only be done once the component has rendered, which not only will drag down performance, but will likely also cause a visible layout change (a la [FOUC](https://en.wikipedia.org/wiki/Flash_of_unstyled_content)) a few hundred microseconds after the initial page render.

I'm not sure how you could do a table or float layout with all the SVG we use for axes.  SVG unfortunately doesn't support any kind of flow layout.  It might be possible to do a table layout using the SVG `foreignObject` but that's not supported by IE.   It might be possible to pull off a table layout if we limit SVG usage to the chart body only, and have the axes labels as non-svg elements.

Ultimately, complex SVG layouts in React is a hard problem, so reducing the use of SVG to the chart body only might be a wise move (indeed, this is why I have the chart title and key in HTML).

parshap: > I'm not sure how you could do a table or float layout with all the SVG we use for axes. SVG unfortunately doesn't support any kind of flow layout.

We'd only need CSS layout to get the legend (a `div`) to flow alongside of fluid-width `svg` elements.

The overall approach taken in this PR is solid and has achieved fluid-width charts for us (@samdalton and I), the implementation just currently has a bug in `LegendChart`.

yang-wei:
Hey @parshap @samdalton, I would really love to merge this but did any of you find a way to implement this in `LegendChart`

samdalton:
@yang-wei I've been using this with a custom legend, and a table-layout to contain the chart and legend side by side. I can update LegendChart to use this same technique, but I haven't currently due to not using it.

samdalton:
@yang-wei @esbullington I've updated this branch with the `LegendChart` updated to use a table cell layout to scale. I've updated the other relevant legend charts to fully support the ViewBoxMixin and did a bit of clean up on old width calculation logic that is no longer necessary. The docs are updated to show this and it seems to work pretty well!

You'll notice in the example I'm giving the chart a fixed height but a 100% width, this is to keep one dimension consistent, and scale it down on the y axis when resizing.

Please let me know if you have any comments/questions on this approach, but it essentially mirrors what we've got going in production with our fork.

yang-wei:
LGTM. Will be merged in #224 since there are some conflict. Thanks again @samdalton.

We might add this into BarChart too when it starts to support multi-series as asked in gitter.

PR #218, opened 2015-05-22
Added props.colors and props.colorAccessor support to BarChart

MatthewHerbst:
Modified `barchart/BarChart` and `barchart/Dataseries` to support the `props.colors` and `props.colorAccessor` functions.

Also passed missing properties into the containing `Chart`.

Note: this is not a breaking change. Users who currently pass the `fill` property to `barchart/BarChart` can still do so, though it will be ignored.

### DIFF
--- a/src/barchart/BarChart.jsx
+++ b/src/barchart/BarChart.jsx
@@ -57,7 +57,17 @@ module.exports = React.createClass({
     var trans = `translate(${ margins.left },${ margins.top })`;
 
     return (
-      <Chart width={props.width} height={props.height} title={props.title}>
+      <Chart
+        viewBox={props.viewBox}
+        legend={props.legend}
+        data={data}
+        margins={props.margins}
+        colors={props.colors}
+        colorAccessor={props.colorAccessor}
+        width={props.width}
+        height={props.height}
+        title={props.title}
+      >
         <g transform={trans} className='rd3-barchart'>
           <DataSeries
             values={values}
--- a/src/barchart/BarChart.jsx
+++ b/src/barchart/BarChart.jsx
@@ -60,7 +60,7 @@ module.exports = React.createClass({
       <Chart
         viewBox={props.viewBox}
         legend={props.legend}
-        data={data}
+        data={props.data}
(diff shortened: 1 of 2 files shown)


src/barchart/BarChart.jsx
```diff
@@ -57,9 +55,19 @@ module.exports = React.createClass({
...
+
     return (
-      <Chart width={props.width} height={props.height} title={props.title}>
+      <Chart
+        viewBox={props.viewBox}
+        legend={props.legend}
+        data={props.data}
+        margins={props.margins}
+        colors={props.colors}
+        colorAccessor={props.colorAccessor}
```
yang-wei: Why both `colors` and `colorAccessor` are passed into Chart ?

MatthewHerbst: Because the `Chart` draws the `Legend`

yang-wei: That means we have to do the samething for other charts as well right.

MatthewHerbst: I thought I had done it already when I first added `colorAccessor` to all the charts. I only hadn't done it to `BarChart` because at that time `BarChart` wasn't even using `colors`

yang-wei: Yes but not all charts, for e.g [PieChart](https://github.com/esbullington/react-d3/blob/master/src/piechart/PieChart.jsx#L49), right? Or it's been overrided?

MatthewHerbst: I think I just forgot to add it to `PieChart`, thanks for pointing that out. I'll go through all of them tomorrow morning and make sure they all have it where missing and send our a pr with those changes.

yang-wei: Thanks. Also [candlestick chart](https://github.com/esbullington/react-d3/blob/master/src%2Fcandlestick%2FCandlestickChart.jsx#L106)

MatthewHerbst: I'm going to make a separate pr for those, so please feel free to merge this one.

PR #221, opened 2015-05-25
Animation for line chart

yang-wei:
Using same approach with scatter chart, that means we do not
use immstruct anymore


src/linechart/LineChart.jsx
```diff
@@ -61,86 +47,71 @@ module.exports = React.createClass({
...
     return (
       <Chart
         viewBox={props.viewBox}
         legend={props.legend}
-        data={data}
+        data={props.data}
         margins={props.margins}
         colors={props.colors}
-        colorAccessor={props.colorAccessor}
+        colorAccessor={(d, i) => d}
```
MatthewHerbst: Why are you forcing the default `colorAccessor` here? You should pass the same one that is coming from props...

src/linechart/LineChart.jsx
```diff
@@ -61,86 +47,71 @@ module.exports = React.createClass({
...
             xScale={scales.xScale}
             yScale={scales.yScale}
+            xAccessor={props.xAccessor}
+            yAccessor={props.yAccessor}
+            hoverAnimation={props.hoverAnimation}
+            circleRadius={props.circleRadius}
+            data={props.data}
+            value={allValues}
+            colors={props.colors}
+            colorAccessor={(d, i) => d}
```
MatthewHerbst: Again, why are you forcing the default `colorAccessor` here? You should pass the same one that is coming from props...

src/linechart/LineChart.jsx
```diff
@@ -61,86 +47,71 @@ module.exports = React.createClass({
...
             width={innerWidth}
             height={innerHeight}
-          /> : <g/> }
+          />
           <XAxis
-            xAxisClassName='rd3-linechart-xaxis'
-            tickFormatting={props.xAxisFormatter}
+            xAxisClassName="rd3-linechart-xaxis"
+            strokeWidth="1"
```
MatthewHerbst: We should never hard code values like this. Make it a property so users can override it if they want.

src/linechart/VoronoiCircleContainer.jsx
```diff
@@ -0,0 +1,76 @@
...
+  },
+
+  getInitialState() {
+    return { 
+      circleRadius: this.props.circleRadius,
+      circleFill: this.props.circleFill
+    }
+
+  },
+
```
MatthewHerbst: This suffers from the same bug that I had previously fixed. You need to also use

```javascript
componentWillReceiveProps(nextProps) {
   this.setState({
      circleRadius: nextProps.circleRadius,
      circleColor:  nextProps.circleFill
   });
},
```

Otherwise it won't update properly. See #211

src/linechart/VoronoiCircleContainer.jsx
```diff
@@ -0,0 +1,76 @@
...
+            cx={props.cx}
+            cy={props.cy}
+            circleRadius={this.state.circleRadius}
+            circleFill={this.state.circleFill}
+        />
+      </g>
+    );
+  },
+
+  _animateCircle() {
```
MatthewHerbst: I again think these numbers should be props, so the user has control over the animation (not to mention that again I don't actually think we should do animation).

src/linechart/DataSeries.jsx
```diff
@@ -12,129 +11,90 @@ module.exports = React.createClass({
 
   propTypes: {
     data: React.PropTypes.array,
-    seriesName: React.PropTypes.string,
-    interpolationType: React.PropTypes.string,
     fill: React.PropTypes.string,
     xAccessor: React.PropTypes.func,
     yAccessor: React.PropTypes.func,
-    displayDataPoints: React.PropTypes.bool
+    colorAccessor: React.PropTypes.func,
```
MatthewHerbst: You are missing `colors: React.PropTypes.func`

src/linechart/DataSeries.jsx
```diff
@@ -12,129 +11,90 @@ module.exports = React.createClass({
...
-      circles = props.data.map(function(point, idx) {
-        var cx, cy;
-        if (this._isDate(point, xAccessor)) {
-          cx = props.xScale(xAccessor(point).getTime());
-        } else {
-          cx = props.xScale(xAccessor(point));
-        }
-        if (this._isDate(point, yAccessor)) {
-          cy = props.yScale(yAccessor(point).getTime());
+        if (this._isDate(props.data[0].values[0], xAccessor)) {
```
MatthewHerbst: Since you removed the check above

```javascript
// Check to see if there is any data to draw lines/circles with
// Return null if there is nothing to draw (null will become <noscript>)
if (props.data.length === 0) {
     return null;
}

this is now going to cause a breaking nullPointerException if `props.data` is empty, which it is by default...

(You need to pay more attention to what you are removing.)

File: src/linechart/DataSeries.jsx [hunk 8]
```diff
@@ -12,129 +11,90 @@ module.exports = React.createClass({
...
+        cy = props.yScale(yAccessor(point).getTime());
+      } else {
+        cy = props.yScale(yAccessor(point));
+      }
+
+      circleFill = props.colors(props.colorAccessor(vnode.point.seriesName, idx));
+      
+      return (
+          <VoronoiCircleContainer 
+              key={idx} circleFill={circleFill} vnode={vnode}
```
MatthewHerbst: Have we stopped splitting props by line?

src/linechart/LineChart.jsx
```diff
@@ -21,35 +14,28 @@ module.exports = React.createClass({
...
-    var structure = immstruct('lineChart', { voronoi: {}, voronoiSeries: {}});
+  render() {
 
     var props = this.props;
 
-    var data = props.data;
-
-    var interpolationType = props.interpolationType || (props.interpolate ? 'cardinal' : 'linear');
+    if (this.props.data && this.props.data.length < 1) {
+      return <g></g>;
```
MatthewHerbst: Is there a reason you choose `<g></g>` over ` null`?

src/linechart/DataSeries.jsx
```diff
@@ -12,129 +11,90 @@ module.exports = React.createClass({
...
+    colorAccessor: React.PropTypes.func,
+    interpolationType: React.PropTypes.string
   },
 
   getDefaultProps() {
     return {
       data: [],
-      seriesName: "",
-      interpolationType: 'linear',
       fill: '#fff',
```
MatthewHerbst: The `fill` prop is no longer needed since we are using `colors` and `colorAccessor`

src/linechart/DataSeries.jsx
```diff
@@ -12,129 +11,90 @@ module.exports = React.createClass({
 
   propTypes: {
     data: React.PropTypes.array,
-    seriesName: React.PropTypes.string,
-    interpolationType: React.PropTypes.string,
     fill: React.PropTypes.string,
```
MatthewHerbst: The `fill` props is no longer needed since we are using `colors` and `colorAccessor`

src/linechart/DataSeries.jsx
```diff
@@ -12,129 +11,90 @@ module.exports = React.createClass({
...
-
+    var xScale = props.xScale;
+    var yScale = props.yScale;
     var xAccessor = props.xAccessor,
-        yAccessor = props.yAccessor;
-
-    // Create array of paths, which we'll map over
-    // to generate SVG lines
+        yAccessor = props.yAccessor,
+        cx, cy, circleFill;
```
MatthewHerbst: Create `cx, cy, circleFill` down where they are actually assigned.

src/linechart/DataSeries.jsx
```diff
@@ -12,129 +11,90 @@ module.exports = React.createClass({
...
-            cx={cx}
-            cy={cy}
-            r={props.circleRadius}
-            fill={props.fill}
-            key={idx}
-            id={props.seriesName + '-' + idx}
+    var lines = props.data.map((series, idx) => {
+      return <Line 
+        path={interpolatePath(series.values)}
+        stroke={props.colors(props.colorAccessor(series.name, idx))}
```
MatthewHerbst: `series.name` here is incorrect. It should take the entire series (`series`). That way users can pass additional information into the series and use it to create colors, or do whatever else they want.

src/linechart/DataSeries.jsx
```diff
@@ -12,129 +11,90 @@ module.exports = React.createClass({
...
+        seriesName={series.seriesName}
```
MatthewHerbst: Pretty sure it should be `seriesName={series.name}` unless we are changing our standards for what should be passed by the users into the `data` array.

src/linechart/DataSeries.jsx
```diff
@@ -12,129 +11,90 @@ module.exports = React.createClass({
...
+      } else {
+        cx = props.xScale(xAccessor(point));
+      }
+      if (Object.prototype.toString.call(yAccessor(point)) === '[object Date]') {
+        cy = props.yScale(yAccessor(point).getTime());
+      } else {
+        cy = props.yScale(yAccessor(point));
+      }
+
+      circleFill = props.colors(props.colorAccessor(vnode.point.seriesName, idx));
```
MatthewHerbst: Again, `vnode.point.seriesName` here is incorrect. It should take the entire series (`vnode`). That way users can pass additional information into the series and use it to create colors, or do whatever else they want.

src/linechart/LineChart.jsx
```diff
@@ -21,35 +14,28 @@ module.exports = React.createClass({
...
-  },
+    margins:        React.PropTypes.object,
+    circleRadius:   React.PropTypes.number,
+    hoverAnimation: React.PropTypes.bool
+ },
 
   getDefaultProps() {
     return {
-      margins:           {top: 10, right: 20, bottom: 40, left: 45},
-      className:         'rd3-linechart',
```
MatthewHerbst: We don't like the `className` anymore?

src/linechart/LineChart.jsx
```diff
@@ -61,86 +47,71 @@ module.exports = React.createClass({
...
         height={props.height}
-        title={props.title}
-      >
-        <g transform={trans} className={props.className}>
-          {dataSeriesArray}
-          {props.hoverAnimation ? <Voronoi
-            structure={structure}
-            data={allValues}
+        title={props.title}>
+        <g transform={trans} className='rd3-linechart'>
```
MatthewHerbst: The `className` should go back to being defined as a default prop. There should never be random hard-coded values, and, it is certainly reasonable that a user might want to override a `className` at some point.

src/linechart/VoronoiCircle.jsx
```diff
@@ -0,0 +1,38 @@
+'use strict';
+
+var React = require('react');
+var d3 = require('d3');
+
+module.exports = React.createClass({
+
+  displayName: 'VoronoiCircle',
+
```
MatthewHerbst: Missing `propTypes` definitions

src/linechart/VoronoiCircle.jsx
```diff
@@ -0,0 +1,38 @@
...
+          d={this.props.voronoiPath} 
+        />
+        <circle
+          onMouseOver={this.props.handleMouseOver}
+          onMouseLeave={this.props.handleMouseLeave}
+          cx={this.props.cx}
+          cy={this.props.cy}
+          r={this.props.circleRadius}
+          fill={this.props.circleFill}
+          className="rd3-linechart-circle"
```
MatthewHerbst: I again think the `className` should be a default prop and not hardcoded.

src/linechart/VoronoiCircleContainer.jsx
```diff
@@ -0,0 +1,76 @@
...
+
+var React = require('react');
+var d3 = require('d3');
+var shade = require('../utils').shade;
+var VoronoiCircle = require('./VoronoiCircle');
+
+module.exports = React.createClass({
+
+  displayName: 'VornoiCircleContainer',
+
```
MatthewHerbst: Missing `propTypes`

src/linechart/VoronoiCircleContainer.jsx
```diff
@@ -0,0 +1,76 @@
...
+    if(props.hoverAnimation) {
+      handleMouseOver = this._animateCircle;
+      handleMouseLeave = this._restoreCircle;
+    } else {
+      handleMouseOver = handleMouseLeave = null;
+    }
+
+    return (
+      <g>
+        <VoronoiCircle
```
MatthewHerbst: No `className` ?

src/linechart/VoronoiCircleContainer.jsx
```diff
@@ -0,0 +1,76 @@
...
+  },
+
+  _restoreCircle() {
+    this.setState({ 
+      circleRadius: this.props.circleRadius,
+      circleFill: this.props.circleFill
+    });
+  },
+
+  _drawPath: function(d) {
```
MatthewHerbst: Shouldn't we be using `d3` to create the path?

MatthewHerbst:
This pr needs some serious work IMHO before it should be merged.

src/linechart/LineChart.jsx
yang-wei: Many of the codes in here are copied from `ScatterChart`. True `null` would be better.

src/linechart/VoronoiCircleContainer.jsx
yang-wei: ok @MatthewHerbst

yang-wei:
@MatthewHerbst Thank you for your time for reviewing this pr. Most of the code came from `scatterChart`. 

Reason of hardcoding those code is to prevent too many of props passing down. You know we are passing props from `Chart` -> `DataSeries` -> `Container` -> `svg element` right? If we depends props for everything we have to concern about them. 

IMO props like `className` are good to hard code. Maybe instead of doing 
```js
<Line className="rd3-line..." />
```
We could do
```js
getDefaultProps() {
  className: "rd3-line"
},
render() {
<Line className={props.className} />
}
```

MatthewHerbst:
That's exactly what I think should happen with `className` and other existing "magic constants".

Regarding props passing: I know it's a pain right now. We do need to figure out a way to do it nicely, but I really do think it is something that should be done. I think a good way of doing it for example, would be to do something like:

```javascript
<LineChart
   data={myData}
   colorAccessor={myColorAccessor}
   ...
   lineProps={myLineProps}
   circleProps={myCircleProps}
/>
```

That way, people still have full access to every property, but, we only unwrap the containers at the component that actually uses them.

MatthewHerbst:
Please see #228 for the changes that need to be made to this.

src/linechart/VoronoiCircleContainer.jsx
yang-wei: We are no longer depends on immstruct to do animation. Will it still caused problem if this isn't added?

MatthewHerbst: Yes - if the user tries to update a component property after an animation has occurred and this isn't added, that property won't get updated.

MatthewHerbst:
Hey, just wondering, should `VoronoiCircleContainer` and `VoronoiCircle ` maybe be pulled out into common since they are used by more than one chart?

yang-wei:
@MatthewHerbst Absolutely. Just trying to figure out the best way of their structure

Issue #225, opened 2015-06-02
className prop for BasicChart


jeans11:
Hi guys !

When we use **BarChart**, we can't specify className prop for **BasicChart** render. Indeed, there is a class for svg element but there isn't class for div element which contain h4 title and svg chart.

MatthewHerbst:
I'll have a pr with these changes + a bit more soon.

MatthewHerbst:
@jeans11 If you look at #227 , while I don't allow you to pass in the className for every sub-component (it would mean adding a massive number of top-level props which we are trying to avoid right now), every Chart, Axis, Title, and Label component will now have a unique className that you can access. That should solve anything you need to do - let me know if you see anything missing.

jeans11:
@MatthewHerbst, thanks !!! I will test it !!!

jeans11:
Hi,

@MatthewHerbst, I can finally test this ! To retrieve the changes, should I change the source of the npm package ?

MatthewHerbst:
The code in the npm package is actually going to be the compiled js if I'm not wrong. Assuming you are using something like `browserify` and `reactify` to compile your JSX into JS, you should probably be alright doing a straight out replace of the code. You might have to do it outside the npm package though, not sure. Let me know how it goes!

jeans11:
Hi,

Thanks @MatthewHerbst. So, I have clone this repo in a thirdparty folder. I use webpack. I meet a little mistake. I think, a dependency is forgotten in the file **LegendChart**  because in line **31**, there is:

```javascript
colors: d3.scale.category20c(),
```
But there is not ```require('d3')```

MatthewHerbst:
Hi @jeans11 . You are completely right, it was accidentally forgotten there - show you how poor our testing currently is :(

For your code to work you can add it in.

Do you want to make a pull-request for the fix? Always nice to have more contributors :+1:

yang-wei:
just curious why there are no errors for this issue

MatthewHerbst:
Because all the charts pass in colors to `Chart` which passes it to `LegendChart`. Since they all pass them in, the `defaultProps.colors` was never getting used, and we don't have a test case for it explicitly, so it probably is failing in the background and we just never noticed.

jeans11:
Hi guys !!

[...]

PR #227, opened 2015-06-03
Changes and additions to Chart components

MatthewHerbst:
This stems from #225 , but it goes a bit farther, but also doesn't do everything the OP in #225 wanted.

1. This adds a `className` to every chart component. `className` is only exposed to the parent of the component - it would be too ugly right now to expose all the class names on the top-level chart, and, the names are unique enough that they should make due for basically all users; I will document the names at some point.
2. **Important: breaking change**: as part of going through all the Chart props, I also went through and **removed supporting `data` being an object - it MUST be an array now**. I think this change is worth it. It makes the code a lot simpler, and allows us to not have to do gimmicky checks and casts which may or may not work. If the user can't figure out that they need to wrap their object(s) in an array, they have much bigger problems - we can only babysit so much.
3. As you can see in this pr, and in #228 , I have started organizing props alphabetically. I think this should be done for every component. It makes props much easier to find and maintain.
4. I added missing `propTypes` and `getDefaultProps` where needed
5. I started the process of converting number-sounding-properties that are actually strings to be numbers. For example, `strokeWidth` in `Label`. While it needs to be a string, intuitively it seems like a number should be supplied. So, I made it so that it takes a number and converts it to a string when actually used.
6. I converted es6 string templates to normal es5 string concatenations. Most places in the code use concatenations, and imho, the templates are a real eye-sore and most contributors won't even know what they are - took me 30 minutes to figure it out since I was thinking someone had accidentally dumped some PHP code into the library, or that $ was doing dereferencing that I didn't know was possible.

### DIFF
--- a/src/common/Legend.jsx
+++ b/src/common/Legend.jsx
@@ -8,19 +8,23 @@ module.exports = React.createClass({
   displayName: 'Legend',
 
   propTypes: {
-    width:         React.PropTypes.number,
-    height:        React.PropTypes.number,
+    className:     React.PropTypes.string,
+    colors:        React.PropTypes.func,
+    colorAccessor: React.PropTypes.func,
+    data:          React.PropTypes.array.isRequired,
+    itemClassName: React.PropTypes.string,
     margins:       React.PropTypes.object,
     text:          React.PropTypes.string,
-    colors:        React.PropTypes.func,
-    colorAccessor: React.PropTypes.func
+    width:         React.PropTypes.number.isRequired
   },
 
   getDefaultProps: function() {
     return {
-      text:          "#000",
+      className:    'rd3-legend',
       colors:        d3.scale.category20c(),
       colorAccessor: (d, idx) => idx,
+      itemClassName: 'rd3-legend-item',
+      text:          '#000'
     };
   },
 
@@ -46,8 +50,9 @@ module.exports = React.createClass({
 
       legendItems.push(
         <li
-          style={itemStyle}
           key={idx}
+          className={props.itemClassName}
+          style={itemStyle}
         >
           <span
             style={textStyle}
@@ -70,7 +75,14 @@ module.exports = React.createClass({
       'listStylePosition': 'inside'
     };
 
-    return <ul style={legendBlockStyle}>{legendItems}</ul>;
+    return (
+      <ul
+        className={props.className}
+        style={legendBlockStyle}
+      >
+        {legendItems}
+      </ul>
+    );
   }
 
 });
--- a/src/common/axes/Label.jsx
+++ b/src/common/axes/Label.jsx
@@ -7,46 +7,76 @@ module.exports = React.createClass({
 
   displayName: 'Label',
 
+  propTypes: {
+    height:              React.PropTypes.number,
+    horizontalTransform: React.PropTypes.string,
+    label:               React.PropTypes.string.isRequired,
+    width:               React.PropTypes.number,
+    strokeWidth:         React.PropTypes.number,
+    textAnchor:          React.PropTypes.string,
+    verticalTransform:   React.PropTypes.string
+  },
+
+  getDefaultProps() {
+    return {
+      horizontalTransform: 'rotate(270)',
+      strokeWidth:         0.01,
+      textAnchor:          'middle',
+      verticalTransform:   'rotate(0)'
+    };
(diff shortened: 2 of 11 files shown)


yang-wei:
Hey @MatthewHerbst thanks for the pr. Really like how you organize code. I would like to merge this.

Just one thing.

> I converted es6 string templates to normal es5 string concatenations.

I would like to keep the es6 style and apply changes to places where it's not used. I believe if you go through our codebase, you will also see a lot of `map` or `foreach` are using `function` instead of `=>` aka fat arrow as their callback. I wish I have time to change them. 

Instead, I am planning to bring in `Babel` and our code will be more `PHP-ish` that time. (`class` will be used instead React.createClass()).Hey @MatthewHerbst thanks for the pr. Really like how you organize code. I would like to merge this.

Just one thing.

> I converted es6 string templates to normal es5 string concatenations.

I would like to keep the es6 style and apply changes to places where it's not used. I believe if you go through our codebase, you will also see a lot of `map` or `foreach` are using `function` instead of `=>` aka fat arrow as their callback. I wish I have time to change them. 

Instead, I am planning to bring in `Babel` and our code will be more `PHP-ish` that time. (`class` will be used instead React.createClass()).

MatthewHerbst:
@yang-wei thanks :)

I'm don't mind moving back to the es6 string templates if that's what you really want. 

What are the advantages of Babel though? I worry that we are spending too much time worrying about non-React things versus just making the library better. It would suck to change the code to use Babel and then a few months from now drop it and have to refactor the code all over again.

yang-wei:
We won't switch to babel until our code reached certain stability which let us refactor easily. 

Moving to babel means we use many es6 feature like `const` and `es6 module` etc. Although currently `reactify` works fine for us but I want to support as much as es6 features. 

Ya I really prefer es6 string template, thanks!

MatthewHerbst:
Alright, I'll make the switch back to templates tomorrow morning, no problem.

Yeah, it is a little disappointing regarding the limited coverage of `reactify`. I've come across that in my own projects. I wonder what their road map looks like.

One thing to think about regarding es6: I visited Uber a few weeks back and spoke with a manager on their front-end team. They don't use any es6 because when it gets compiled, it's extremely difficult if not impossible to debug through stack-traces in the console. For that reason they don't use any compiled js, don't allow any anonymous functions, and don't even use JSX.

yang-wei:
> They don't use any es6 because when it gets compiled, it's extremely difficult if not impossible to debug through stack-traces in the console.

Thanks for the note. Ya there must be some trade off but so far I never come across any difficult debug problem. Maybe our code base is still small so I think it's fine.

MatthewHerbst:
pr has been updated to keep the es6 string templates.

yang-wei:
Just wondering why these 2 commits look similar

MatthewHerbst:
@yang-wei I forgot to add any files to first commit but didn't want to deal with squashing them - the last time I messed with that I seriously broke some stuff on the git branch and @esbullington had to fix it.

yang-wei:
One solution is to delete your remote branch, squash them and push again. That works for me.

MatthewHerbst:
@yang-wei would you mind giving me the commands for that? I don't want to loose all the work if I mess up.

yang-wei:
Doing `git log` you will see something like this:
```
* 8af72a4 (HEAD, MatthewHerbst-MatthewHerbst-225/Chart-props) Reverted to es6 string templates
* 39ef427 Reverted back to es6 string templates
* 84da135 Made sure all Chart related components had a className....
```
Because we want to squash `8af72a4` and `39ef427` into a commit. So you will do
```
git rebase -i 39ef427

// you will see
pick 39ef427 Reverted back to es6 string templates
pick 8af72a4 Reverted to es6 string templates
 
# Rebase 39ef427..8af72a4 onto 39ef427 (       1 TODO item(s))
#
```
Just change the word `pick` into `squash` or `s`.
```
pick 39ef427 Reverted back to es6 string templates
squash 8af72a4 Reverted to es6 string templates
```
You will see another window show up showing combination of 2 commit messages. Just change them and it will be squashed. Then you can delete the remote branch and repush.

However this time I will merge this later if you don't have time to do this. You pushed some great thing, thanks.

MatthewHerbst:
Done :)

PR #228, opened 2015-06-03
Refactoring of ScatterChart code + minor changes to utils.flattenData

MatthewHerbst:
This pr does some major refactoring to the `ScatterChart` code.

1. Fixes missing props, `propTypes`, and `getDefaultProps`
2. Fixes the bug where animation was making the component fail to respond to changes due to missing `componentWillReceiveProps`
3. Fixes non-standard calls to `colorAccessor`
4. Because data for `ScatterChart` goes through `utils.flattenData`, that method had to be modified to ensure that each record still retained all original data, and that each record knew from which series it came from.
5. `utils.flattenData` had a small potential bug fixed, replacing `in` with `hasOwnProperty`
6. Code across `ScatterChart` has been simplified
7. Based on #227 , support for `data` as an Object has been removed
8. "Magic constants" have been moved into `defaultProps`
9. Tests have been updated to support the above
10. es6 string templates were removed in favor of traditional es5 string concatenations. As I note in #227 , this is for standardization of the code base as well as maintainability

@yang-wei this is what the pr you submitted in #221 should look like. These changes are all the comments I made on that pr, plus some.

### DIFF
--- a/src/scatterchart/DataSeries.jsx
+++ b/src/scatterchart/DataSeries.jsx
@@ -9,60 +9,65 @@ module.exports = React.createClass({
   displayName: 'DataSeries',
 
   propTypes: {
-    data: React.PropTypes.array,
-    fill: React.PropTypes.string,
-    xAccessor: React.PropTypes.func,
-    yAccessor: React.PropTypes.func,
-    colorAccessor: React.PropTypes.func
+    circleRadius:  React.PropTypes.number.isRequired,
+    colors:        React.propTypes.func.isRequired,
+    colorAccessor: React.PropTypes.func.isRequired,
+    data:          React.PropTypes.array.isRequired,
+    height:        React.PropTypes.number.isRequired,
+    xAccessor:     React.PropTypes.func.isRequired,
+    xScale:        React.PropTypes.func.isRequired,
+    yAccessor:     React.PropTypes.func.isRequired,
+    yScale:        React.PropTypes.func.isRequired
   },
 
   getDefaultProps() {
     return {
-      data: [],
-      fill: '#fff',
-      xAccessor: (d) => d.x,
-      yAccessor: (d) => d.y
+      data: []
     };
   },
-  
+
   render: function() {
-    var props = this.props;
-    var xScale = props.xScale;
-    var yScale = props.yScale;
-    var xAccessor = props.xAccessor,
-        yAccessor = props.yAccessor,
-        cx, cy, circleFill;
+    var props     = this.props;
+    var xScale    = props.xScale;
+    var yScale    = props.yScale;
+    var xAccessor = props.xAccessor;
+    var yAccessor = props.yAccessor;
 
     var voronoi = d3.geom.voronoi()
       .x(function(d){ return xScale(d.coord.x); })
       .y(function(d){ return yScale(d.coord.y); })
       .clipExtent([[0, 0], [ props.width , props.height]]);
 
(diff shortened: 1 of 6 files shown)


yang-wei:
LGTM. Especially your changes in utils.flattenData, it made me easier to use `colorAccessor` in `LineChart`.

Just will really like to keep es6 string templates.

MatthewHerbst:
@yang-wei :) you should have seen how ugly my hack for that way before I realized I could just add it like I did here, haha.

See #227 for the es6 templates, that way we can keep that conversation in one thread.

yang-wei:
@MatthewHerbst I successful solve this problem by using series name.
```js
// set this in CartesianChartMixins
getDefaultProps() {
  colorAccesors: (d, i) => d
}

// then in legend, line etc
var comp = data.map((series, idx) {
  colors(accessor( series.seriesName, idx ))
})

```

This way we can avoid inserting new prop (seriesIndex) in flattendata utils.

What do you think? @MatthewHerbst

MatthewHerbst:
@yang-wei that doesn't solve it, it breaks it.

The whole point of `colorAccessor` is that it allows the user to give additional data to their data elements and then base the color off of that data. For example:

For example, where before the `data` array looked like this:

```json
[
  {
    "name": "series1",
    "values": [
      {"x": 10, "y": 20}, 
      {"x": 20, "y": 30}
    ]
  },
  {
    "name": "series2",
    "values": [
       {"x": 10, "y": 40}, 
       {"x": 20, "y": 60}
    ]
  },
]
```

With this, the only parameters available to `colorAccessor` are what you have, the `seriesName` and the series `idx`, which is very limiting.

Using `colorAccessor` as I designed it however, passing `d` and the series `idx`, you can now have a `data` structure such as:

```json
[
  {
    "name": "series1",
    "values": [
      {"x": 10, "y": 20, "uuid": "34539d66-f036-462d-a27f-284c84217485"}, 
      {"x": 20, "y": 30, "uuid": "046fd3aa-900e-4054-8e20-1d97f6e1a209"}
    ]
  },
  {
    "name": "series2",
    "values": [
       {"x": 10, "y": 40, "uuid": "7397fc14-1ca7-4b70-806e-458153d371e9"}, 
       {"x": 20, "y": 60, "uuid": "afcfa65c-83e5-4e8f-b746-6199d191a75f"}
    ]
  },
]
```

With the above `data` structure, I can pass in my own `colorAccessor` that looks like:

```javascript
(d, idx) => d.uuid
```

This allow the user to base the color on literally whatever they want, without having to do weird math hacks to match the data elements in the chart back to the original data they have. I use this in a production environment right now. It is absolutely needed. I'm not sure why you would want to limit the amount of data available for the user to base decisions on...

MatthewHerbst:
You don't know what the user will want to access in their passed-in functions, so don't limit the data available to them.

MatthewHerbst:
@yang-wei I made a mistake with the above. Update coming soon. You made me realize a bug in `colorAccessor`

MatthewHerbst:
Maybe to make things simpler, could you merge this pr and #227 and then I'll make a new pr for the `colorAccessor` change? That change is going to touch every chart so I think it needs its own pr.

MatthewHerbst:
To be clear on what I'm thinking, I've realized that `colorAccessor` needs one more parameter.

Instead of:

```javascript
(series, idx) => idx
```

It should be

```javascript
(series, d, idx) => idx  // Where d is the current element within the series
```

PR #232, opened 2015-06-15
Update packagejson

yang-wei:
Non breaking changes

 * remove immstruct dependency
 * update dev dependencies at package.json
 * run lint and eliminate those warnings


MatthewHerbst:
LGTM!

PR #233, opened 2015-06-24
Exposed stroke-width and added stroke-dasharray support to line chart

samdalton:
I took this approach to adding `stroke-width` and `stroke-dasharray` props to LineChart. It is different from how we specify colors with a `colorAccessor` prop but seemed vastly simpler.

screen shot 2015-06-24 at 10 52 09 am

### DIFF
--- a/docs/examples/main.js
+++ b/docs/examples/main.js
@@ -51,7 +51,9 @@ var Demos = React.createClass({
     var lineData = [
       { 
         name: 'series1',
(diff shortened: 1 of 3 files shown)


MatthewHerbst:
LGTM.

What do you mean it is simpler than `colorAccessor`? Just seems like a straight props pass to me.

samdalton:
@MatthewHerbst `colorAccessor

PR #235, opened 2015-07-11
fixes toString bug with single-digit RGB values in utils.shade

transcranial:
In utils.shade, if the values of red, green, or blue are single-digit, the resulting string may not be a proper hex string. Single digits need to be prepended with '0'.


yang-wei:
@transcranial thanks for the pr but I think we want to leverage [d3 color method(]https://github.com/mbostock/d3/wiki/Colors) instead of maintaining our owns. 

Then we can support other color formats too =)

esbullington:
@transcranial thank you for the PR.

@yang-wei does this interfere with the d3 color method somehow?

yang-wei:
@esbullington No. Feel free to merge

PR #236, opened 2015-07-16
Add optional props for BarChart

jeans11:
Hi guys !!!

In some cases, data from API are shared between diffrent components. In addition, the data coming from API have their own structure. So, I was wondering if it would be wise to setup labels and values that we wish to convey to the components.

### DIFF
--- a/src/barchart/BarChart.jsx
+++ b/src/barchart/BarChart.jsx
@@ -25,6 +25,8 @@ module.exports = React.createClass({
 
   getDefaultProps() {
     return {
+			nameLabelProp: "label",
+			nameValueProp: "value",
       yAxisTickCount: 4,
       margins: {top: 10, right: 20, bottom: 40, left: 45},
       hoverAnimation: true
@@ -35,9 +37,9 @@ module.exports = React.createClass({
 
     var props = this.props;
 
-    var values = props.data.map( (item) => item.value );
+    var values = props.data.map( (item) => item[props.nameValueProp] );
 
-    var labels = props.data.map( (item) => item.label );
+    var labels = props.data.map( (item) => item[props.nameLabelProp] );
 
     var margins = props.margins;
 


MatthewHerbst:
If we want to go down this route, it would be worth doing it for all charts at once, not just `BarChart`

jeans11:
Yes, absolutely !

yang-wei:
@jeans11 Thanks for the PR but we really don't want to just do `BarChart`.

If you could do for other charts, then I will update demo page and documentation.

Thanks

jeans11:
Ok, I will try to do it for other charts in the coming days.

jeans11:
Hi guys !

I have modified **AreaChart**, **PieChart**, and **Treemap**. For other charts, can I update __flattenData__ function of utils module ?

For example:

```javascript
exports.flattenData = (data, propMapping, xAccessor, yAccessor) {
...
}
```

With __propMapping__ is an object with:

```javascript
{
    label: nameLabelProp,
    value: nameLabelValue
}
```

MatthewHerbst:
You need to modify it for the other charts to get these changes, correct? So, go for it!

yang-wei:
Hi @jeans11, I am sorry to tell you that we decided to make `Bar Chart` into multi series chart ( #245 ).
So instead of doing it this way, what do you think if we make these props accessor function.

So instead of 
```
getDefaultProps() {
  nameLabelProp: "label",
  nameValueProp: "value",
}
```
The accessor way will be
```
getDefaultProps() {
  labelsAccessor: d => d.labels,
  valuesAccessor: d =>  d.values,
}
// to use it
render() {
 var { data, labelsAccessor, valuesAccessor } = this.props;
 var labels = valuesAccessor(data);
 var values = valuesAccessor(data);
}
```
What do you think ?

jeans11:
Hi guy !!!

Oh yes, good idea to make Bar chart to series chart !!!

@yang-wei, I update all chart with props accessor function, it's better ^^.

PR #242, opened 2015-07-26
Use babelify for es6 and jsx transformation

### LINKED ISSUE #174: Support es6 by babeljs

yang-wei:
fix #174


esbullington:
Sounds good to me.  My intention was to move to babelify whenever the React community did, and it sounds like that's well underway now.

Issue #237, opened 2015-07-18
ColorAccessor's future


MatthewHerbst:
So, I've been giving more thought to colors and our graphs, and come to this realization: we either need multiple` colorAccessor`s for different parts of the graph, or, each `*Chart` needs to define its own `colorAccessor`.

Let me explain:

The current design goal (not what it currently is) of `colorAccessor` is:

```javascript
colorAccessor: (series, d, seriesIdx) => seriesIdx
```

That looks fine, except it falls apart if you try to use it for both say, `AreaChart` and `ScatterChart`. That because the `AreaChart` (or `LineChart`) both have something a ScatterChart doesn't have - lines. This is important, because right now the `colorAccessor` assumes that it will have access to each datum (datapoint in the `values` array. However, that is not how lines work. Lines are made up of two points, and thus, to properly color a line segment, `colorAccessor` would need access to both points that makes up that segment.

In addition, the way we implement colors right now is somewhat limiting. For example, I might want to create a `LineChart` where segments above a certain value are one colors, and segments below that value are another. This is not currently possible. It is also not currently possible to say, create a horizontal gradient `AreaChart`.

So, to the solutions:

**1)** We could continue to use a single `colorAccessor` and simply null out values that we can't give you. This option is by far the easiest to implement, though would require documentation (and the user to look at said documentation and/or examples) to know which data parts were available.

**2)** We could move `colorAccessor` out of the common mixin and implement it within each `*Chart`. This is the next easiest solutions, and allows the user full control over colors with minimal code changes from us. However, the downside to this is more code to maintain, and, over time if we are not careful it could become very different to color one chart from the next.

**3)** We could create three common accessors. If you think about any `*Chart`, they are all made up of the same things: lines and/or points and/or areas. Thus, if we had a common color function to handle each, we could keep them in the mixin yet every `*Chart` would still have full access to all color features. This is a slightly bigger code change, but I am happy to do it. The three accessors would be (obviously, some `*Chart` would not use all three:
```javascript
// Where d is the single data point that defines the point
pointColorAccessor: (series, seriesIdx, d, dIdx) => seriesIdx

// Where p0 and p1 are the two points (each with values x,y) that define the line segment
lineSegementColorAccessor(series, seriesIdx, p0, p0Idx, p1, p1Idx) => seriesIdx

// Where p0 and p1 are the two points (each with values x,y,y0) that define the area segment
[...]

Issue #219, opened 2015-05-22
*Chart and DataSeries standardization


MatthewHerbst:
**tl;dr**: we need to choose a single format for creating Charts and their elements via the `DataSeries`.

Currently, *Charts and their respective DataSeries are implemented one of two ways:

1) The *Chart creates a `dataSeriesArray` by mapping over the data. Each `DataSeries` produces one chart element, such as a single line (with its circles) in `LineChart`

Ex. from `LineChart`:

```javascript
...
render() { 
    var dataSeriesArray = data.map( (series, idx) => {
        return (
            <DataSeries
              key={idx}
              structure={structure}
              xScale={scales.xScale}
              yScale={scales.yScale}
              seriesName={series.name}
              data={series.values}
              width={innerWidth}
              height={innerHeight}
              fill={props.colors(props.colorAccessor(series, idx))}
              circleRadius={props.circleRadius}
              xAccessor={props.xAccessor}
              yAccessor={props.yAccessor}
              interpolationType={interpolationType}
              displayDataPoints={props.displayDataPoints}
            />
        );
    });

    return (
        <Chart
            ...
        >
            <g
                {dataSeriesArray}
            >
            ...
        </Chart>
    );
}
```

2) The *Chart has a `DataSeries` element which creates all the chart elements. The single `DataSeries` creates all the elements by mapping over them, such as all the bars in `BarChart`.

Ex. from `BarChart`:

```javascript
...
render() {
    return (
        <Chart
            viewBox={props.viewBox}
            legend={props.legend}
            data={props.data}
            margins={props.margins}
            colors={props.colors}
            colorAccessor={props.colorAccessor}
            width={props.width}
            height={props.height}
            title={props.title}
        >
            <g transform={trans} className='rd3-barchart'>
                <DataSeries
                    values={values}
                    labels={labels}
                    yScale={yScale}
                    xScale={yScale}
                    margins={margins}
                    data={props.data}
                    width={props.width - sideMargins}
                    height={props.height - topBottomMargins}
                    colors={props.colors}
                    colorAccessor={props.colorAccessor}
                    hoverAnimation={props.hoverAnimation}
                />
                ...
            </g>
        </Chart>
     );
}
```

We should chose one way of doing things to be standard. This will allow for much simpler code reading, more generic code plugins such as through Mixins, and easier maintenance.

Personally, I prefer the first option. To me, each `DataSeries` should infer a single series of data such as a line or a bar, rather than all of the series of data.

[...]

PR #247, opened 2015-08-07
Data point symbols

nightlyop:
Currently, all data points in line charts are displayed as circles. I need the possibility to use other symbols (eg. squares, stars, diamonds, "v", ...) to display the data points. Moreover I need to be able to specify the symbol for every data point. The reason therefore is, that my chart has to use the same system as an older one we're selling. In our case the kind of symbol indicates special observations made when the displayed value was measured.

I tried to make the Voronoi Container more generic and so far the symbols `circle` (default; same as before) and `rect` can be used.

`VoronoiContainer.jsx` replaces `VoronoiCircleContainer.jsx` because most of the symbols I'd like to create will share most of the code in here. Because of that I also renamed variables like `circleFill` to `symbolFill`.

What do you think about this approach?


yang-wei:
@nightlyop Hi Chris thanks for the pr. Briefly looked at them and I like what you are adding to this library.

However first of all I don't really prefer `symbol` as the name. For me `marker` sounds like a better name. What do you think ?

nightlyop:
@yang-wei You're right, `marker` is better. I'll change that.

yang-wei:
> VoronoiContainer.jsx replaces VoronoiCircleContainer.jsx
agree.

But I saw code replication between VoronoiCircle and VoronoiRect. VoronoiContainer should not worry about what to rendering. It's job is to handle state(in our case).

So I will prefer the `switch` statement to go into VoronoiChild( currently VoronoiCircle ) and the voronoi child will decide what to render. Does this make sense?

nightlyop:
Do you mean a component VoronoiContainer and a component VoronoiMarker and VoronoiMarker contains a `switch` to decide what to render? That would sound good for me as well. I'm a beginner and don't know what's better but I'm glad if I can improve it with your feedback ;-)

MatthewHerbst:
I like this!

@yang-wei I actually think the switch statement is in the correct place. A VoronoiChild such as `VoronoiCircle` should have no knowledge of the other children such as `VoronoiRect`.

yang-wei:
@nightlyop Yes VoronoiMarker will be a better name. 

What I see about VoronoiRect is it has a lot of code duplication with VoronoiCircle. That's what I want to avoid so I think the VoronoiMarker should contains the switch statement.

nightlyop:
Ok, I'll rename everything and replace `symbol` by `marker`
After reading 
* [Smart and Dumb Components](https://medium.com/@dan_abramov/smart-and-dumb-components-7ca2f9a7c7d0)
* [Container Components](https://medium.com/@learnreact/container-components-c0e67432e005)

I also thinkt the switch statement is at the correct place. The components `VoronoiRect` and `VoronoiCircle` are indeed very similar. But I expect other kind of markers, for which there is no ready to use SVG component (eg. a star or a "v"), to be more different.

yang-wei:
IMHO what make VoronoiRect and VoronoiCircle similar is the `<path>` component.
```
      <g>
        <path
          onMouseOver={this.props.handleMouseOver}
          onMouseLeave={this.props.handleMouseLeave}
          fill='white'
          d={this.props.voronoiPath} 
        />
       ...
      </g>
```
Because they are voronoi-ed so we need this component on every child of container component. Imagine that if we create another marker then we have one more `VoronoiNewMarker` file. 
Then one if we have to change our `<path>` api, we need to touch a lot of files. This is what I want to avoid. Let me know what do you think :smile:

nightlyop:
How about putting that part in a mixin?

yang-wei:
I don't see mixin is an alternative but if you don't mind to explain, can you elaborate?

nightlyop:
A collegue told me that mixin is not a good solution because they won't be supported in the future. How about making a higher order component wrapping `path` (according to [Mixins Are Dead. Long Live Composition](https://medium.com/@dan_abramov/mixins-are-dead-long-live-higher-order-components-94a0d2f9e750))

yang-wei:
No one knows the future of mixin but if you look at [react-future](https://github.com/reactjs/react-future/blob/master/01%20-%20Core%2F02%20-%20Mixins.js) or [decorator](http://raganwald.com/2015/06/26/decorators-in-es7.html) it's still possible to use mixin.

HIgher order component is the thing I've been thinking all the time but I still can come out with a good pattern.

nightlyop:
To summarize the update:
- I renamed everything from symbol to marker
- I made a component `VoronoiArea` containing `path` (we discussed making a mixin but I like this solution)
- I created a marker in the form of a star
- The form of the marker can be defined for each data point seperately. If less markers are specified than there are data points, the last specified marker is used for the remaining data points (eg if only one marker is specified, this marker is used for all data points). However, if multiple series share a data point, some points might get the wrong marker. The other properties of the marker (eg size) can be set per series. I'm not happy with the code structure in `DataSeries.jsx` because I'd like to be able to set all properties for each data point seperately. But I didn't figure out a good solution. It would be necessary to access the attributes of each data point when creating `VoronoiContainer`. Although some further work is necessary, it is working.

MatthewHerbst:
Should these changes be applied to ScatterChart as well?

nightlyop:
I think so but unfortunately I don't have time for that. I'm trying to add a possibility to display a grid now.

MatthewHerbst:
Mmm. And I'm working on #237 . Maybe I'll have the time to make these changes to ScatterChart after I finish that. And I agree, the differing structures of `DataSeries` bothers me as well. We had some discussion about it in #219 but haven't talked about it in a while.

nightlyop:
I saw that discussion after my PR. However I think it is quite a lot of work to figure out a good architecture and it might be better if someone with more experience (and time) could do that somewhen.

.gitignore
```diff
@@ -19,3 +19,6 @@ build/
 
 # OSX
 .DS_Store
+dist/public/js/react-d3.min.js
+dist/public/js/react-d3.min.js.map
+test.js
```
yang-wei: what's `test.js` for. If you wish to ignore these file I think it's better to do it in another pr.

nightlyop: ugh, I was a little confused. I thought that's from the master. Don't know where that's from. I'll remove it.

nightlyop:
The different kind markers can be used for scattercharts as well. I haven't touched the legend so far and I don't think I'll have time therefore in the next weeks. But if you merge the code I can write a help page.

yang-wei:
:+1 will look at these later. need more time since it had so many files change

nightlyop:
Sure, I'm not in a hurry.

nightlyop:
Might have been better to make a new branch for this fix.. Sorry. Shall I make one?

yang-wei:
yes ! If you could please squash the bugfix commit. thanks

nightlyop:
Aaaw, sorry for the git mess. When I noticed that I forgot the `ScatterChart` I tried to amend the changes to the first commit but it didn't work as I expected. :-/

nightlyop:
I wrote https://github.com/esbullington/react-d3/wiki/Marker about what I made in this branch. It's

yang-wei:
@nightlyop Thanks for the hard work but I really think that our code base will become complicated after merging this. Give me some time, I might do a little bit refactoring before merging this into master.

src/common/marker/index.js
```diff
@@ -0,0 +1,11 @@
+/**
+ * Created by cfaes on 24.08.2015.
```
yang-wei: Is your id cfaes ? We don't give credit in our code base. We do it in changelog or other place. Is it ok for me to remove this?

yang-wei:
See issue #261

src/common/marker/index.js
nightlyop: Sure. That's added by webstorm automatically and I forgot to remove it.

yang-wei:
Will close this due to inactive.

PR #250, opened 2015-08-21
Grid

nightlyop:
Grid lines can be plotted in all charts having an axis. The grid lines are part of the component `AxisTicks` because a grid line has the same starting position as a tick. Grid lines are not plotted at positions which have an `AxisLine`. To display the chart correctly, the grid lines are plotted behind/before the data points.

Have a nice weekend!


MatthewHerbst:
Looks awesome!

yang-wei:
LGTM to me. Just 2 thing:
 * please solve the conflict since i merge #251 (I merged #251 first because it's small and easier to look at )
 * could you add documentation [here](https://github.com/esbullington/react-d3/wiki/Axis) regarding to grid =)

nightlyop:
Ok. I'll write some documentation and will have a look for merge conflicts with my other branch (#247).

yang-wei:
Thanks. 
Awesome man ! @nightlyop

yang-wei:
I will add related test soon :smile:

MatthewHerbst:
I've found a problem with this: the `AxisTicks` component has in it's `PropTypes`:

```javascript
orient: React.PropTypes.oneOf(['top','bottom','left','right']).isRequired,
orient2nd: React.PropTypes.oneOf(['top','bottom','left','right']),
```

However, if you look at the actual `XAxis` and `YAxis` components, you will see that both have in their `PropTypes`:

```javascript
xOrient: React.PropTypes.oneOf(['top', 'bottom']),
yOrient: React.PropTypes.oneOf(['left', 'right']),
```

I am very confused about the use of `orient` and `orient2nd`. Is there a reason they are not called `xOrient` and `yOrient` respectively? I'm assuming it's because when you pass the props for the `YAxis`, they are the reverse of when you pass for the `XAxis`. I feel like this could be better be handled with an addition prop being a boolean determining the type of axis that the `AxisTicks` is being used for.

nightlyop:
`AxisTicks` is used universally and can have one of the four orientations `['top','bottom','left','right']`. The same is true for a second axis (if there is one). If you draw grid lines with `AxisTicks` you have to know the orientation of both axes (to know where the two axes cross). Then it is possible to prevent drawing a grid line at the same position where the second axis is drawn. It is not a very good solution because it might be possible to have more than two axes one day.

How I understand your idea: With a boolean you would know the type of the axis (`XAxis` or `YAxis`) but you wouldn't know the orientation of the second axis. So lets assume you would know that you are drawing an `XAxis` and the second axis is a `YAxis`. But you would not know if the `YAxis` crosses the `XAxis` at the start (`left`) or end (`right`). You would not know which vertical grid line has to be omitted.

MatthewHerbst:
Ah, I think I understand a little more now.

Is there a reason the grid can't just be drawn first so that if an axis does exist in the same place as a grid line, that line would just be covered by the axis's line?

nightlyop:
You're welcome! That was my first try because it would be the simplest solution. But:
* it works only if the grid lines are not thicker than the axis line
* it didn't work in one case. If I remember correctly, it didn't work when the Y axis was on the right side. But i'm not sure anymore and I don't know why it didn't work.

MatthewHerbst:
Hmm. I feel like the first case isn't something to worry about. If the user makes the grid-lines larger than the axis line, that is their choice and they can easily correct it.

I'm interested to know more about the second case. Maybe I'll try it out this coming weekend and see what I can find.

nightlyop:
If it works it would be the better solution. I don't know why it didn't work in one case...

PR #251, opened 2015-08-21
InterpolationType for LineChart

nightlyop:
Pass property `interpolationType` from `LineChart` to `DataSeries` to actually being able to use it.


MatthewHerbst:
+1

PR #253, opened 2015-08-24
Fix problem with typescript caused by #741de83

nightlyop:
Could we add this lines to package.json as a temporary solution until somebody has time to solve the problem as suggested by @yang-wei:
> One way to solve this is to compiled all code(babel, jsx) from src into directory. Then ts user can use our module straight from there.


yang-wei:
Sure. Can you let me know how does this solve typescript user problem?

nightlyop:
Actually I don't fully understand it. In our project I have to add `babelify` and then (with the added code to `package.json`) I can compile react-d3 as vendor library when building my project. I think without the changes gulp tries to use reactify or tsify to compile but that does not work.

yang-wei:
Maybe it's related to [this](https://github.com/babel/babelify#why-arent-files-in-node_modules-being-transformed). Will merge this

PR #254, opened 2015-08-28
Bugfix: Don't add h4 title element to chart if there is no title

nightlyop:
Although I didn't set a title for the chart, there was the h4 `rd3-chart-title`
element in my DOM. In my case `this.props.title = ""` if I don't set a title.
However, this is not the case for phantom.js. Does that depend on the
browser version? I'm using Chrome and Firefox. Anyway, with the changes I
don't have the title element and the karma tests end successfully.


yang-wei:
LGTM !

PR #255, opened 2015-09-01
Set axis line width of line charts

nightlyop:
Pass xAxisStrokeWidth and yAxisStrokeWidth to being able to set the width of the line axis of line charts.


MatthewHerbst:
+1 I was trying to figure out why it was "none" haha

PR #259, opened 2015-09-11
fix viewBoxObject in the example

vladikoff:
The code above uses {{ but the example docs are not


yang-wei:
Nice catch. Thanks !

PR #266, opened 2015-10-01
param hoverAnimation fixed

RassaLibre:
This param did not make it into VoronoiCircleContainer so setting this param was totally useless. But I fixed it.


yang-wei:
thanks !

PR #276, opened 2015-11-12
Add .editorconfig and some minor code cleanup

mwhite:
I am implementing horizontal orientation for cartesian charts.  These are ongoing refactorings in service of that.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,3 @@
+[*.{js,jsx}]
+indent_style = space
+indent_size = 2
--- a/src/areachart/AreaChart.jsx
+++ b/src/areachart/AreaChart.jsx
@@ -38,7 +38,7 @@ module.exports = React.createClass({
 
     var interpolationType = props.interpolationType || (props.interpolate ? 'cardinal' : 'linear');
 
-    var {innerWidth, innerHeight} = this.getDimensions();
+    var {innerWidth, innerHeight, trans} = this.getDimensions();
 
     if (!Array.isArray(data)) {
       data = [data];
@@ -82,8 +82,6 @@ module.exports = React.createClass({
       .values((d)=> { return d.values; });
 
     var layers = stack(data);
-
-    var trans = `translate(${ props.margins.left },${ props.margins.top })`;
 
     var dataSeries = layers.map( (d, idx) => {
       return (
--- a/src/areachart/AreaChart.jsx
+++ b/src/areachart/AreaChart.jsx
@@ -38,10 +38,7 @@ module.exports = React.createClass({
 
     var interpolationType = props.interpolationType || (props.interpolate ? 'cardinal' : 'linear');
 
-    // Calculate inner chart dimensions
-    var innerWidth, innerHeight;
-    innerWidth = this.getOuterDimensions().width - props.margins.left - props.margins.right;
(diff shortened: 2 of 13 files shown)


yang-wei:
@mwhite code looks good to me but the test is failing. Please fix it and I will merge.

mwhite:
It's a problem with the test. Nodename is expected to be uppercase G but is actually lowercase g in whatever version of PhantomJS is being run.  Not sure when this was introduced.

I was intending to fix it by applying toUppercase() or the like before the comparison, is that okay?

yang-wei:
@mwhite Is it working at your local? I pull your pr and tests are green locally.

MatthewHerbst:
Please remove the `.editorconfig` from this pr. Everyone uses different editors, and this file is not source code.

mwhite:
It's a standard that works with many different editors. http://editorconfig.org/

There is still a bug that needs to be fixed, one of the refactorings was not quite correct.

MatthewHerbst:
@mwhite it doesn't matter. It's not source code, it doesn't belong in the repo. Coding standards can be enforced by documenting them and ensuring them while reviewing PRs.

parshap:
> Please remove the .editorconfig from this pr. Everyone uses different editors, and this file is not source code.

> @mwhite it doesn't matter. It's not source code, it doesn't belong in the repo. Coding standards can be enforced by documenting them and ensuring them while reviewing PRs.

I disagree. Coding standards are best enforced with code. EditorConfig is a great way to help people use the right whitespace settings. `package.json` isn't source code either, but it helps developers have the right versions of dependencies.

MatthewHerbst:
@parshap `package.json` is required to build the source code. EditorConfig is great, but it's not standard. Next I'll put my `.eslint` file is, and someone else will put their `.jshint`, and it gets out of control. There is a reason you never see these types of files in any major application.

mwhite:
I think you will find quite a high percentage of popular projects have .eslintrc and many have .editorconfig.

MatthewHerbst:
The way I see it: if it's not source code or part of the build process, it doesn't belong. Now, I have plenty of projects where I actually run a lint job as part of the build project, and so yeah,I'l have an `.eslintrc` file in there. However, I wouldn't include one in the repo in the hopes that other people might have ESLint installed in their editors that will pick up my file just to enforce standards.

So, if we are going to include style standard files, I think we should include a task as part of the build to enforce those standards. That way we are not liable to each developers local system preferences. That's all I've got to say about this.

parshap:
@MatthewHerbst: That's fair. I think @mwhite is implicitly asking this project to adopt EditorConfig as a standard — not just haphazardly adding his own configuration file.

> So, if we are going to include style standard files, I think we should include a task as part of the build to enforce those standards. That way we are not liable to each developers local system preferences.

That'd be nice. I think `.editorconfig` in itself is useful even without a build task, but I can see your perspective too.

parshap:
Also: The code changes here are good. Let's not get caught up with EditorConfig and throw the baby out with the bathwater.

@yang-wei: Any other changes required to merge this?

mwhite:
I'm adding tests.  Also, I opened a different PR.

yang-wei:
@parshap I am good with these changes.

Just @mwhite now added some commits to #277 so I might need extra time to have a look at it.

PR #277, opened 2015-11-13
Horizontal cartesian charts

mwhite:
Some description of the implementation here: https://github.com/esbullington/react-d3/commit/abdd34fa00bc652b37b4be720ac99523c481c555

Feedback welcome.

There are no tests, I suppose it wouldn't be hard to test the rotation and translation behavior of axis and tick labels, but at least this is open to review now.

**In the images below, the less-than-ideal axis label placements and the candlestick bottom axis overlap issue are due to the fact that these screenshots are of the example page with no properties modified other than horizontal={true}.**  When adjusting `margins` and (for candlestick) `yAxisOffset` to better values for horizontal charts, they look better.  To implement two different default margins depending on orientation would be additional work.

screen shot 2015-11-13 at 3 45 48 pm
screen shot 2015-11-13 at 3 46 04 pm
screen shot 2015-11-13 at 3 46 11 pm
screen shot 2015-11-13 at 3 46 17 pm
screen shot 2015-11-13 at 3 46 24 pm

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,3 @@
+[*.{js,jsx}]
+indent_style = space
+indent_size = 2
--- a/src/areachart/AreaChart.jsx
+++ b/src/areachart/AreaChart.jsx
@@ -38,7 +38,7 @@ module.exports = React.createClass({
 
     var interpolationType = props.interpolationType || (props.interpolate ? 'cardinal' : 'linear');
 
-    var {innerWidth, innerHeight} = this.getDimensions();
+    var {innerWidth, innerHeight, trans} = this.getDimensions();
 
     if (!Array.isArray(data)) {
       data = [data];
@@ -82,8 +82,6 @@ module.exports = React.createClass({
       .values((d)=> { return d.values; });
 
     var layers = stack(data);
-
-    var trans = `translate(${ props.margins.left },${ props.margins.top })`;
 
     var dataSeries = layers.map( (d, idx) => {
       return (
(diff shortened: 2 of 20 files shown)


MatthewHerbst:
This looks great! Not sure I've ever seen a horizontal candlestick chart before!

Isn't a horizontal line chart also achievable just by flipping the data you have on each axis? Regardless, good to have the feature standardized on all charts!

mwhite:
> Isn't a horizontal line chart also achievable just by flipping the data you have on each axis?

Probably, I just wasn't familiar with the data manipulation and thought it would be better to handle it as an abstraction at the SVG level using rotation.  As a side effect of that, you could, say, toggle the horizontal property of a chart and have nothing in the DOM change other than the transform attributes of the top-level container and the text nodes.

mwhite:
To make things orientation-agnostic, it might be good to change yOrient and xOrient to use 'high' and 'low' rather than 'right', 'left', 'top', 'bottom' at some point.

mwhite:
I just tested xOrient='top' manually for the first time, which should put the x-axis on the right for horizontal charts, but it doesn't really work.

It seems to be the same behavior as the current master, so I made an issue: https://github.com/esbullington/react-d3/issues/278

yang-wei:
@mwhite This feature looks good to me, awesome ! 

I will look at the code again tomorrow. Tests are welcomed but not necessary. Please also add some documentation related to this in our [wiki](https://github.com/esbullington/react-d3/wiki).

PR #279, opened 2015-11-17
Fixed some tests and removed series information which was contained in each data point

nightlyop:


nightlyop:
I don't have time to fix the rest of the tests but there are not a lot left to fix...

src/utils/index.js
```diff
@@ -117,7 +117,6 @@ exports.flattenData = (data, xAccessor, yAccessor) => {
         },
         d: item,
         id: series.name + j,
-        series: series,
```
MatthewHerbst: This inclusion is intentional and should not be removed. It allows functions iterating over these points to have access to custom props passed to the entire series that are not available at the point level. For example, I might want to color my points based on a custom prop `myColorProp` that I attach to the series object.

nightlyop: Ah ok, so this is intended. I thought it might slow down the charts if you have a lot of points...

Issue #274, opened 2015-11-03
Variable-width charts?


idris:
Is there a way to specify a % for a chart width, or have it fill the entire container?

The height could maintain some aspect ratio.

MatthewHerbst:
I don't think there is a way to pass a percentage. Your best bet would be to create a container around the chart. Then, when you create the chart, pass it the width and height of the container, which will have the effect of filling the container. All charts take width and height (in pixels) parameters.

idris:
Right, but what if the container width is relative to the window?  Basically, I want to use this in a responsive flex-width website.

MatthewHerbst:
Make a pr to support percent :)

nightlyop:
I think it might be nice to have the possibility to have dynamic chart size. So far I'm using a workaround (copied from gitter):

1. the chart is embedded in div1
2. I set the states graphWidth and graphHeight in the constructor of the class/page to an initial size.
3. I create the chart:
```javascript
    return React.createElement(RD3.LineChart, {
                 legend: true,
                 data: lineData,
                 width: this.state.GraphWidth,
                 height: this.state.GraphHeight,
                 //title: 'Line Chart',
                 interpolate: 'cardinal'
                 // hoverAnimation: true    // highlight line when mouseover; doesn't work on touch devices
         });
```

4. 4! (don't know why markup changes my 4 to a 1 -.-) The function resizeGraph() does 
```javascript
   this.setState({graphWidth: div1.clientWidth, graphHeight: div1.clientHeight});
```
 --> state change causes rendering. resizeGraph() is also called in componentDidMount() to update the size of the graph (the initial size might not be correct but I don't know the size of div1 before the first rendering).
5. When there's a resize event, the event handler calls resizeGraph().

vladnicula:
One way to go about it would be to have a window resize event, and publish width and height changes. I was doing that, but with this approach we can't render variable width charts on the server.

What do you guys think about this approach?

MatthewHerbst:
Looking forward to seeing it in action!

yang-wei:
Feel free to pr directly if this issue is solved

PR #280, opened 2015-11-20
allow string and number propTypes for width/height attributes

igl:
width & height can be strings (as seen here: https://reactiva.github.io/react-d3-website/)

(Is this the correct repo to make PRs? Or are you working on the reactiva org now?)


igl:
I tried to run the test locally and thought i am missing some global deps but it looks like your master is broken.

yang-wei:
> I tried to run the test locally and thought i am missing some global deps but it looks like your master is broken.

This is a known issue. This will be merged after #277

yang-wei:
> Is this the correct repo to make PRs? Or are you working on the reactiva org now?

@igl Further notification will be do but now it's fine to open your pr here =)

MatthewHerbst:
I think the example website is wrong. Width and height should be numbers only.

igl:
If they are numbers only, how do you differentiate between a percentage and pixel value? Or em/rem etc.

MatthewHerbst:
@igl That's an excellent point! At the moment the code does not support percentages. However, take a look at #274 where that's being talked about. I assume that with that change then yes, strings would be accepted. Above I was just saying where the code is now.

Issue #261, opened 2015-09-13
marker


yang-wei:
A sneak peek of current implementation. Thanks @nightlyop
Marker only applied to 2 charts:
 * scatter chart
 * line chart

![image](https://cloud.githubusercontent.com/assets/5494874/9835974/c0673900-5a3f-11e5-89f4-0ca1a4a3cdd2.png)

How to use (API):

![image](https://cloud.githubusercontent.com/assets/5494874/9835986/039f5e3c-5a40-11e5-9bbf-a897020c0a2c.png)

Implementation was good but I did not happy with the api because I think data should be pure data. To me, thing like `markerOuterRadius`(from above) seems like a component property.

What api should look like: 
```
// image we have data as picture above
<ScatterChart
  data={scatterData}
  marker={["rect", "star", "circle"]}        <- pass in marker we want 
/>
```
*if marker prop length is less than series length, a default marker will be passed in =) *

A more complex marker?
```
// image we have data as picture above
<ScatterChart
  data={scatterData}
  marker={[{marker: "rect", appearance: { width: 5, height: 5, animationFactor: 1.25}},  ... ]}        <- pass in marker we want as object. appearance is used to describe the marker.
/>
```
Hey  @nightlyop if you have some spare time, would like to know what do you think. You can see the branch I am working at [here](https://github.com/esbullington/react-d3/tree/DataPointSymbols-cleanup)

Todo list:
##### fundamental
 * [x] marker mixin
 * [ ] new api
 * [ ] documentation related to marker
 
##### add on
 * [ ] custom marker

nightlyop:
To pass a separated marker object to set the properties of the markers sounds good to me. I like your example for a more complex marker! However, I would not use mixins (https://medium.com/@dan_abramov/mixins-are-dead-long-live-higher-order-components-94a0d2f9e750 but we discussed that already in another issue).

yang-wei:
@nightlyop Yes I am aware of this. However, I think mixins is really good for code resusablity, I still can't figure out how to replace our mixins with HOC. 

If you look at [react-future repository](https://github.com/reactjs/react-future/blob/master/01%20-%20Core%2F02%20-%20Mixins.js), mixins will be support in other way so I think it's safe enough to use mixins since it makes our codebase simpler =)

PR #286, opened 2015-12-21
New featuers which need some more time before merging

nightlyop:
This is not meant to merge directly but maybe to start a little discussion and/or find someone to make it ready for a merge. We did quite a few things over the last months but right now we don't have time to spend more time. Things we did but are not in `esbullington/react-d3` (yet):
- Created more markers like stars, triangles, polygons, squares...
- Possibility to define minimum and maximum value for an axis (e.g. possibility to define that the starting tick of the X. axis has to be between 0 and 5).
- Possibility to specify the properties of the line (in a line chart) between each marker.

I think this are the main featuers we implemented (because we need it). Unfortunately we can't spend more time right now. But maybe someone likes some stuff we did and then it might be easier to spend some time to improve/clean up code instead of implementing the same feature again. I'm specially not happy with the marker props. Would be nice to pass an Object for each marker to the chart instead of how its done now.

Now there are a lot of props like: `markerName`, `markerWidth`, `markerCorners`, `markerBaseColor`, `markerInnerRadius`. IMHO it would be better to pass an object like
```
marker: {
   name: 'star',
   innerRadius: 5,
   outerRadius: 10,
   ...
}
```
This also applies to the props for a custom grid.

What do you think?

See also the issue #261 about the markers.


MatthewHerbst:
Agree on passing more props wrapped as objects. I'm also interested to know what ya'll think of using `context` - would be useful for passing things through all the layers from user-facing to the bottom.

nightlyop:
I didn't add the possibility to plot a second y axis. I also restructured the props for the axes and grid and replaced the components `XAxis` and `YAxis` by a single component `Axis`. However, that's all for line charts only and I didn't merge it yet (although it should be straight forward): https://github.com/nightlyop/react-d3/tree/multipleYAxes

I've off now but maybe someone wants to spend some time on it during the holidays?

Merry x-mas and happy new year!

yang-wei:
Would be happy if someone willing to take this.

PR #287, opened 2015-12-21
Tooltips for line, scatter, bar and pie charts

jfcaiceo:
This new feature use the `TooltipMixin` to show a tooltip  when `onMouseOver` event is triggered. The charts that use this mixin (`LineChart`, `ScatterChart`, `BarChart` and `PieChart`) now support two new props: `showTooltip` (determines whether to use the tooltip) and `tooltipFormat` (a function that determines the tooltip text).

### DIFF
--- a/src/barchart/BarChart.jsx
+++ b/src/barchart/BarChart.jsx
@@ -5,12 +5,12 @@ var d3 = require('d3');
 var DataSeries = require('./DataSeries');
 var utils = require('../utils');
 
-var { Chart, XAxis, YAxis } = require('../common');
-var { CartesianChartPropsMixin, DefaultAccessorsMixin, ViewBoxMixin } = require('../mixins');
+var { Chart, XAxis, YAxis, Tooltip } = require('../common');
+var { CartesianChartPropsMixin, DefaultAccessorsMixin, ViewBoxMixin, TooltipMixin } = require('../mixins');
 
 module.exports = React.createClass({
 
-  mixins: [ CartesianChartPropsMixin, DefaultAccessorsMixin, ViewBoxMixin ],
+  mixins: [ CartesianChartPropsMixin, DefaultAccessorsMixin, ViewBoxMixin, TooltipMixin ],
 
   displayName: 'BarChart',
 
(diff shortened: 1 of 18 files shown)


MatthewHerbst:
:+1:

akshaymohite:
@jfcaiceo / @MatthewHerbst  couple of inputs -

- As the onMouseOver event is added on path as well (https://github.com/esbullington/react-d3/blob/master/src/linechart/VoronoiCircle.jsx#L21-L22) - the co-ordinates being shown for tooltip are not exactly what we might want

- It would be better if we support having xLabel and yLabel in front of tool tips - it would be needed in case of multiple elements in series (Multiple y lines etc.)

jfcaiceo:
@akshaymohite:
* That's not a problem because I always get the coordinates of the circle inside the path: https://github.com/jfcaiceo/react-d3/blob/master/src/linechart/VoronoiCircleContainer.jsx#L56
* I don't understand your point. You are talking about having the series name in the tooltip?

akshaymohite:
@jfcaiceo:
> That's not a problem because I always get the coordinates of the circle inside the path: https://github.com/jfcaiceo/react-d3/blob/master/src/linechart/VoronoiCircleContainer.jsx#L56

:+1: 

> I don't understand your point. You are talking about having the series name in the tooltip?

Yes, series name and xLabel as well if possible.

yang-wei:
LGTM ! Code looks good to me. Just one thing, theres is a [todo](https://github.com/esbullington/react-d3/pull/287/files#diff-0217eec6176b6f3e938bac402013aed9R19). Are you going to solve it before I merge?

jfcaiceo:
When the tooltip is near the right edge of the chart, it would be good to change the orientation of the tooltip to the left and always were inside the chart. That's the `todo`. I don't think that's a required feature to this pull request. But... what do you think @yang-wei?

BigDataSamuli:
Thanks, this is very useful. However, for the tooltip divs, might be better to only define minimal styles inline and instead add classes for easy customization?

Also, +1 for passing the series name into the tooltip.

Finally, maybe consider officially supporting custom tooltip components, this sort of works already with a tooltipFormat function that returns a component, e.g:

```javascript
tooltipFormat={v => (<div>{"Hello " + v.xValue + ", " + v.yValue}</div>)}
```

jfcaiceo:
@BigDataSamuli:
> However, for the tooltip divs, might be better to only define minimal styles inline and instead add classes for easy customization?

That's the best option I think, but now the library is clean of stylesheets... that's good too. I tried to keep it clean of other assets with this change.

> Also, +1 for passing the series name into the tooltip.

:+1: changed

> Finally, maybe consider officially supporting custom tooltip components, this sort of works already with a tooltipFormat function that returns a component

That's true. I changed the `text` prop, for a `child` prop ( that can be a string, number or react element) to support custom tooltips. https://github.com/jfcaiceo/react-d3/blob/master/src/common/Tooltip.jsx#L10

BigDataSamuli:
@jfcaiceo:

> That's the best option I think, but now the library is clean of stylesheets... that's good too. I tried to keep it clean of other assets with this change.

Yeah, I was thinking you could take most of the inline styles out and the tooltip would still be somewhat usable even without any external stylesheets. In any case, you could add the classes so that one can at least easily override the styles with !important.

Also, new browsers support a CSS property "pointer-events: none", which would be useful here to keep the tooptip itself from interfering with the mouseover events.

yang-wei:
>  I don't think that's a required feature to this pull request. But... what do you think 

Yeah no problem for me.

Aleczhang1992:
How to use toolTip?There are some problems when I add showTooltip and tooltipFormat.It seems doesnt work.my code

 `
<LineChart
          legend={true}
          data={lineData}
          width={'100%'}
          height={400}
          viewBoxObject={{
            x: 0,
            y: 0,
            width: 1000,
            height: 400
          }}
          xAxisTickInterval={{unit: 'day', interval: xInteral}}
          xAxisFormatter={(val)=>Common.getTime_ymd_hms("date",val)}
          showTooltip={true}
          tooltipFormat={()=>"like"}
          gridHorizontal={true}
        />
`

PR #288, opened 2015-12-24
Support negative values for Bar Chart

jfcaiceo:
screen shot 2015-12-24 at 11 02 32 am


yang-wei:
Seems good. 

Although it's not necessary, adding unit test for this will be good =)

PR #293, opened 2016-01-05
Render voronoi geometry on top of linechart lines

BigDataSamuli:
Currently the linechart line paths interfere with the hover animation by triggering mouse out events when the cursor goes on top of a line. This is a simple but effective fix, rendering the voronoi geometry (including the circles) on top of the lines.


BigDataSamuli:
This other pull request https://github.com/esbullington/react-d3/pull/286 contains the same change so closing this.

PR #297, opened 2016-01-21
added support for ordinal axes

napisani:
added support for ordinal axes

### DIFF
--- a/src/utils/index.js
+++ b/src/utils/index.js
@@ -1,9 +1,8 @@
 var d3 = require('d3');
 
-
-Array.prototype.unique = function() {
-  var o = {}, i, l = this.length, r = [];
-  for(i=0; i<l;i+=1) o[this[i]] = this[i];
+var findUniqueValues = function(list) {
+  var o = {}, i, l = list.length, r = [];
+  for(i=0; i<l;i+=1) o[list[i]] = list[i];
   for(i in o) r.push(o[i]);
   return r;
 };
@@ -17,7 +16,7 @@ exports.calculateScales = function(chartWidth, chartHeight, xValues, yValues)  {
     xScale = d3.time.scale()
       .range([0, chartWidth]);
   } else if (xValues.length > 0 && Object.prototype.toString.call(xValues[0]) === '[object String]'){
-    tickDist = (chartWidth) / xValues.unique().length;
+    tickDist = (chartWidth) / findUniqueValues(xValues).length;
     for (tickRangeNum = chartWidth; tickRangeNum > 0; tickRangeNum -= tickDist) {
       tickRange.push(tickRangeNum);
     }
@@ -33,7 +32,7 @@ exports.calculateScales = function(chartWidth, chartHeight, xValues, yValues)  {
     yScale = d3.time.scale()
       .range([chartHeight, 0]);
   } else if (yValues.length > 0 && Object.prototype.toString.call(yValues[0]) === '[object String]'){
-    tickDist = (chartHeight) / yValues.unique().length;
+    tickDist = (chartHeight) / findUniqueValues(yValues).length;
     for (tickRangeNum = chartHeight ; tickRangeNum > 0; tickRangeNum -= tickDist) {
       tickRange.push(tickRangeNum);
     }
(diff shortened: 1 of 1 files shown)


src/utils/index.js
```diff
@@ -1,14 +1,29 @@
 var d3 = require('d3');
 
 
-exports.calculateScales = (chartWidth, chartHeight, xValues, yValues) => {
+Array.prototype.unique = function() {
+  var o = {}, i, l = this.length, r = [];
+  for(i=0; i<l;i+=1) o[this[i]] = this[i];
+  for(i in o) r.push(o[i]);
+  return r;
+};
```
mwhite: It's generally not good to extend native prototypes with non-standard methods in library code.

napisani:
I changed 'unique' to be a helper function. I'd like to re-request a pull.

src/utils/index.js
```diff
@@ -17,7 +31,14 @@ exports.calculateScales = (chartWidth, chartHeight, xValues, yValues) => {
   if (yValues.length > 0 && Object.prototype.toString.call(yValues[0]) === '[object Date]') {
     yScale = d3.time.scale()
       .range([chartHeight, 0]);
-  } else {
+  } else if (yValues.length > 0 && Object.prototype.toString.call(yValues[0]) === '[object String]'){
+    tickDist = (chartHeight) / findUniqueValues(yValues).length;
+    for (tickRangeNum = chartHeight ; tickRangeNum > 0; tickRangeNum -= tickDist) {
+      tickRange.push(tickRangeNum);
+    }
+    console.log(JSON.stringify(tickRange));
```
yang-wei: I think you forget to remove this line ?

yang-wei:
> added support for ordinal axes

Your pull request did a lot than this. why did you change back es6 syntax to es5 ?

napisani:
updated.

src/utils/index.js
```diff
@@ -1,14 +1,28 @@
...
 
+var findUniqueValues = function(list) {
+  var o = {}, i, l = list.length, r = [];
+  for(i=0; i<l;i+=1) o[list[i]] = list[i];
+  for(i in o) r.push(o[i]);
+  return r;
+};
 
-exports.calculateScales = (chartWidth, chartHeight, xValues, yValues) => {
+exports.calculateScales = function(chartWidth, chartHeight, xValues, yValues)  {
```
yang-wei: indeed you do not have to change these 2 lines.

We prefer using `exports.calculateScales = (chartWidth, chartHeight, xValues, yValues) => {`

src/utils/index.js
```diff
@@ -78,29 +91,17 @@ exports.flattenData = (data, xAccessor, yAccessor) => {
...
-        // in allValues or in xValues or yValues
-        if (isNaN(y)) {
-          return;
-        }
         yValues.push(y);
         yNode = y;
       }
 
-      var xyCoords = `${x}-${yNode}`;
+      var xyCoords = (x + "-" + yNode);
```
yang-wei: I am sure 
```
`${x}-${yNode}`
```
is more readable.

src/utils/index.js
napisani: how about now?

yang-wei: you still miss one @napisani

napisani:
updated

src/utils/index.js
```diff
@@ -78,24 +91,12 @@ exports.flattenData = (data, xAccessor, yAccessor) => {
...
           // as in the case of candlestick, default to y value of 0
           yNode = 0;
         });
       } else {
-        // Check for NaN since d3's Voronoi cannot handle NaN values
-        // Go ahead and Proceed to next iteration since we don't want NaN
-        // in allValues or in xValues or yValues
-        if (isNaN(y)) {
-          return;
-        }
```
yang-wei: is there any reason you remove all the isNan() check. We need this in graph with nested data.

napisani:
yes, I removed the checks because if you are working with an ordinal axis then your data is inherently no a number. For example, a point on a graph of "Time VS Location" may look like this:
{
    x: new Date(),
    y: "New York"
}

For a graph of this nature, all y values will be NaN, but the changes support having strings as Axis values now.

Issue #298, opened 2016-01-25
[Feature] How can i interact with each segment in pie chart by clicking or something like this?


namKolo:
I really want to interact with each segments in pie chart. But in fact, i can't do with this lib.

yang-wei:
Currently not in our roadmap so I will close this

PR #299, opened 2016-01-25
Add handleClickOnSegment handler for pie chart

namKolo:
Interactive with pieChart


namKolo:
when you have time, please review this commit to fix issuse #298

yang-wei:
Add a onClick/onMouseOver handler to this chart is ok with me. I know that it's a requirement to build interactive chart. 

But I can't just merge this. Because only supporting this single chart with onClick handler is not fair. I want to make all charts as universal as possible.

PR #301, opened 2016-01-26
Added VennDiagram

tyfoo:
Added a `VennDiagram` using [venn.js](https://github.com/benfred/venn.js/) (see http://www.benfrederickson.com/venn-diagrams-with-d3.js/ for more details)

### DIFF
--- a/docs/examples/main.js
+++ b/docs/examples/main.js
@@ -5,6 +5,7 @@ var d3 = require('d3');
 var hljs = require("highlight.js");
 var rd3 = require('../../src');
 var BarChart = rd3.BarChart;
+var VennDiagram = rd3.VennDiagram;
 var LineChart = rd3.LineChart;
 var CandlestickChart = rd3.CandlestickChart;
 var PieChart = rd3.PieChart;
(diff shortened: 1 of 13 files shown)


yang-wei:
This is great but I don't think it's the requirement for most case in building chart. So I would choose to avoid adding another layer of dependency into this library.

Anyway thanks !

tyfoo:
Thanks for the quick feedback. 

I understand. Would you like me to do a version without the venn.js dependency?

yang-wei:
>  understand. Would you like me to do a version without the venn.js dependency?

that would be awesome !

tyfoo:
@yang-wei please see amended commit without the `venn.js` dependency. 

I've taken the functions used from venn.js and put them under `venn.js` folder, including credit to @benfred and the original library / blog post. I hope that is okay.

MatthewHerbst:
@yang-wei I would say based on how much code is copy/pasted from `venn.js`, just use the dependency. There is no reason to duplicate code, and, none of us are going to spend as much time maintaining and improving that code as the original author.

PR #303, opened 2016-01-27
Optimize tooltip

jfcaiceo:
Added `shouldComponentUpdate`method to `Chart`. From now on the `Chart` will not call the `render` method when the tooltip state change.


yang-wei:
Good one ! looks good to me !

